PR #49, opened 2021-01-07
Incorporate the third-party licenses.

td202:


abizjak:
Would it not be possible to change the tool so that it produced a bit more readable pages, and much smaller.

We have a small number of different licenses for the dependencies, so it would be more readable to just list those, and then under/above list what dependencies are under that license.

td202:
> 
> 
> Would it not be possible to change the tool so that it produced a bit more readable pages, and much smaller.
> 
> We have a small number of different licenses for the dependencies, so it would be more readable to just list those, and then under/above list what dependencies are under that license.

It depends on the purpose of this page. If it is merely informational, then we can probably cut down what we include. If it is to meet our obligation to include licenses with the distributed software, then I think we should reproduce each license in full, since many of them include the required copyright notice.

limemloh (approved):

PR #56, opened 2021-01-11
Wallet guides

jens-concordium:
The guides for the mobile wallet has been completely remade after all the changes and new features.

It now includes a guide on how to install the app as the first thing. This change has also been mirrored on the Downloads page.

PR #57, opened 2021-01-11
Identities and accounts updates

jens-concordium:
Fixed links and references. A couple of the links will not work before they are merged with the other PRs, though, e.g. the newly added reference to initial accounts.


source/testnet/references/id-accounts.rst
```diff
@@ -1,16 +1,12 @@
...
-.. _Concordium ID: /testnet/docs/downloads#concordium-id
+.. _`anonymity revocation`: /testnet/references/id-accounts.html#revoking-anonymity
+.. _Concordium ID: /testnet/see-also/downloads.html#concordium-id
+.. _`initial account`: /testnet/see-also/glossary.html#initial-account
 .. _Notabene: https://notabene.id/
 .. _Onfido: https://onfido.com/
-.. _mainnet: /testnet/docs/glossary#mainnet
-.. _on-chain: /testnet/docs/glossary#on-chain
-.. _anonymity revocation: #revoking-anonymity
+.. _`mobile wallet`: /test/guides/get-started.html
```
abizjak, line 9: This link does not lead anywhere as far as I can see.

Moreover I don't think  we should use the term "mobile wallet". We use `Concordium ID`

source/testnet/references/id-accounts.rst
```diff
@@ -23,13 +19,13 @@ Identities and accounts
...
 To facilitate compliance with relevant regulations, an account can only be
-created from an *identity* which is issued `off-chain`_ by an Identity provider.
+created from an *identity* which is issued :ref:`glossary-off-chain` by an Identity provider.
 While an account has to be created from an identity, the user's privacy is still
 protected, and the account owner's identity can only be revealed via the process
-of *anonymity revocation*, which can only happen under stringent regulations,
+of `anonymity revocation`_, which can only happen under stringent regulations,
 such as upon a court order. In particular, a key feature of the design of
 identities and accounts is that the identity provider cannot reveal the identity
 of an account on their own.
```
abizjak, line 31: This needs some rework due to initial accounts. The identity provider can tell who they belong to, since it creates them (in essence).

source/testnet/references/id-accounts.rst
```diff
@@ -74,13 +70,14 @@ on the provided documentation.
...
-purposes, and will not be supported on the `mainnet`_.
+purposes, and will not be supported on the :ref:`glossary-mainnet`.
 
 Creating an account
 ===================
 
-Once a user has an identity, they can use it to create a number of accounts. In
-contrast to obtaining an identity, opening an account is an `on-chain`_ action
+Once a user has an identity, they can use it to create a number of accounts, besides the
+`initial account`_ created by the identity provider upon creating the identity. In
```
abizjak, line 79: There does not seem to be a glossary entry for initial account.

abizjak (commented):
Some links need to be fixed, and the first few paragraphs rephrased to accommodate the initial account effects.

source/testnet/references/id-accounts.rst
jens-concordium: The link target is added in https://github.com/Concordium/concordium.github.io/pull/56. 

I'll change it to Concordium ID.

source/testnet/references/id-accounts.rst
jens-concordium: Yeah, that makes sense. I'll try reworking it.

source/testnet/references/id-accounts.rst
jens-concordium: This entry is also added in another PR. (https://github.com/Concordium/concordium.github.io/pull/56)

source/testnet/references/id-accounts.rst
jens-concordium: I tried reworking this bit, but I'm not sure if it's sufficient. If not, I might need some back-up from someone who knows more about it.

source/testnet/references/id-accounts.rst
```diff
@@ -23,16 +18,21 @@ Identities and accounts
...
-created from an *identity* which is issued `off-chain`_ by an Identity provider.
-While an account has to be created from an identity, the user's privacy is still
-protected, and the account owner's identity can only be revealed via the process
-of *anonymity revocation*, which can only happen under stringent regulations,
-such as upon a court order. In particular, a key feature of the design of
-identities and accounts is that the identity provider cannot reveal the identity
-of an account on their own.
+send, receive :ref:`glossary-GTU`) a user must create an *account*.
+
+The user will get an `initial account`_ at the same time as an *identity* has been issued
```
abizjak, line 23: Shouldn't this be an internal link
```
:ref:`initial account <glossary-initial-account>`
```

source/testnet/references/id-accounts.rst
```diff
@@ -23,16 +18,21 @@ Identities and accounts
...
+by an identity provider. As the initial account is submitted to the chain by the
+identity provider, said identity provider will know the owner of the initial account. For this
+reason, the user may consider not using the initial account, and create a regular account instead.
+
+The owner of a regular account, will only be known by the owner himself. To facilitate
```
abizjak, line 28:
```suggestion
Regular accounts' owner is not known to the identity providers, or any other single entity. To facilitate
```

source/testnet/references/id-accounts.rst
```diff
@@ -74,13 +74,14 @@ on the provided documentation.
...
-purposes, and will not be supported on the `mainnet`_.
+purposes, and will not be supported on the :ref:`glossary-mainnet`.
 
 Creating an account
 ===================
 
-Once a user has an identity, they can use it to create a number of accounts. In
-contrast to obtaining an identity, opening an account is an `on-chain`_ action
+Once a user has an identity, they can use it to create a number of accounts, besides the
+`initial account`_ created by the identity provider upon creating the identity. In
```
abizjak, line 83: And an internal link here as well.

abizjak (approved):
I'd like the two minor things fixed, but after that we should merge this.

PR #66, opened 2021-01-12
Add windows instructions

abizjak:


make.bat
```diff
@@ -28,6 +30,14 @@ if errorlevel 9009 (
 %SPHINXBUILD% -M %1 %SOURCEDIR% %BUILDDIR% %SPHINXOPTS%
 goto end
 
+:dev
+sphinx-autobuild.exe %SOURCEDIR% %BUILDDIR/html -E
+goto end
+
+:lint
+doc8 ./source --ignore D004 --ignore D002
```
limemloh, line 38: should use the config `doc8.ini` found in the root of this project

```suggestion
doc8 ./source --config ./doc8.ini

```

limemloh (approved):

make.bat
abizjak: Right, but this won't work because of the way line endings are handled by git on windows (in the default setup).

Unless we want to add D004 and D002 to ignored list, always.

limemloh: I see.
But if the above succeeds on windows, it must be using the doc8.ini already, since it defaults to max-line-length of 79 characters, so this is fine then.

abizjak: I think it did succeed, but I cannot check again.

Reading this https://pypi.org/project/doc8/ it looksl like it will  use doc8.ini in CWD if it finds it.

PR #67, opened 2021-01-13
Move topbar navigation to sidebar

limemloh:
Since we currently only have two items in the topbar and it is taking a lot of space.

Instead the items should be in the sidebar:

image

### DIFF
--- a/source/_templates/layout.html
+++ b/source/_templates/layout.html
@@ -7,10 +7,11 @@
 <link href="{{ pathto("_static/style.css", True) }}" rel="stylesheet" type="text/css">
 {% endblock %}
 
-{# Add the topbar #}
-{% block extrabody %}
-    <nav class="topbar">
-        <a href="{{ pathto('smart-contracts/index') }}" class="{{ 'active' if entry is defined and entry == 'smart-contracts/index' }}"> <i class="icon fa fa-file-o"></i> Smart Contracts </a>
-        <a href="{{ pathto('testnet/index') }}" class="{{ 'active' if entry is defined and entry == 'testnet/index' else '' }}" ><i class="icon fa fa-connectdevelop"></i> Testnet</a>
-    </nav>
+{% block sidebartitle %}
+{{super()}}
+<nav class="sidebar-top-menu">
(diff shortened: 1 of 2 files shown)


source/_templates/layout.html
```diff
@@ -7,10 +7,11 @@
...
-    <nav class="topbar">
-        <a href="{{ pathto('smart-contracts/index') }}" class="{{ 'active' if entry is defined and entry == 'smart-contracts/index' }}"> <i class="icon fa fa-file-o"></i> Smart Contracts </a>
-        <a href="{{ pathto('testnet/index') }}" class="{{ 'active' if entry is defined and entry == 'testnet/index' else '' }}" ><i class="icon fa fa-connectdevelop"></i> Testnet</a>
-    </nav>
+{% block sidebartitle %}
+{{super()}}
+<nav class="sidebar-top-menu">
+    <a href="{{ pathto('testnet/index') }}" class="{{ 'active' if entry is defined and entry == 'testnet/index' else '' }}" ><i class="icon fa fa-connectdevelop"></i> Testnet</a>
+    <a href="{{ pathto('smart-contracts/index') }}" class="{{ 'active' if entry is defined and entry == 'smart-contracts/index' }}"> <i class="icon fa fa-file-o"></i> Smart Contracts </a>
+</nav>
```
abizjak, line 15: In the future I'd like to see some principled approach to this kind of change. What is the purpose, what is the overall plan.

For example, I think this change makes it less clear to discover there is the smart contracts tab as well because it is less prominent on the page. So it is unclear to me whether we should merge it.

limemloh: Ideally we would move towards the final design of the documentation, but to my knowledge there is still no plan for how the final documentation should be structured, 
and this PR is just another approach, which I find better than the current topbar with two items.

jens-concordium: To me it makes sense to move the tabs to the sidebar, but I agree that the inactive tab is a little too invisible now.

It might work better if the tabs had the same 'look' as they do now, but extruding from the sidebar instead of the main content area.

limemloh: Something like this:

image

jens-concordium: Yep, exactly!

abizjak (approved):
Assuming it is the new style that you proposed, let's get this merged.

And please update the PR description with the new screenshot.

PR #72, opened 2021-01-13
Explanation on how to find out current/past exchange rates

amaurremi:


abizjak (approved):

abizjak:
I think the explanation is good, but the linter should be satisfied before this is merged.

PR #84, opened 2021-02-10
fixed typo in winning probability in glossary

abhconcordium:
From memory, this should be `1-(1-f)^a`, not `1-(1-f)a`.

Could someone verify?

### DIFF
--- a/source/testnet/see-also/glossary.rst
+++ b/source/testnet/see-also/glossary.rst
@@ -438,5 +438,5 @@ Winning probability
 ===================
 
 The winning probability is the probability that a baker wins in a given slot.
-The probability is *1-(1-f)^α*, where *f* is the difficulty parameter and *α* is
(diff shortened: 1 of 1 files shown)


abhconcordium:
@limemloh I wasn't sure if there was a proper way of doing exponentiation in this file format

limemloh:
> @limemloh I wasn't sure if there was a proper way of doing exponentiation in this file format

We can enable a plugin to allow math in the form of latex expressions, but if this is the only place, then just use the `html` directive.

abizjak (changes requested):
Good catch, but can you please make it pretty as well (with the exponential).

abhconcordium:
> Good catch, but can you please make it pretty as well (with the exponential).

Latest push renders like this

image

limemloh (approved):

limemloh:
@abhconcordium 
Because of the recent changes, you should also fix this in the `testnet4` branch

PR #85, opened 2021-02-10
Reflect refactor of add -> name

abhconcordium:
update documentation to reflect refactor of config account add to config account name, as well as the name parameter not being optional.

### DIFF
--- a/source/testnet/references/concordium-client.rst
+++ b/source/testnet/references/concordium-client.rst
@@ -196,7 +196,7 @@ Add accounts and keys
 See also the :ref:`Managing accounts<managing_accounts>` section.
 
 Add named account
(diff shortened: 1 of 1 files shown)


abhconcordium:
Tied to CB-990

limemloh (approved):

abhconcordium:
After discussion with @limemloh, set as draft to avoid merging into external documentation, while feature is still internal only.

abizjak:
We need to remember that this is incompatible with the latest released version, so we must not deploy this version onto the website.

limemloh:
Now that only the `testnet4` branch is deployed, this should be safe to merge into `main`

PR #91, opened 2021-03-16
Reject reasons

Bargsteen:
- Update code examples
- Add guide about returning custom errors
- Minor cleanup

### DIFF
--- a/source/smart-contracts/guides/custom-errors.rst
+++ b/source/smart-contracts/guides/custom-errors.rst
@@ -1,0 +1,55 @@
+.. _custom-errors:
+
+====================
+Return custom errors
+====================
+
+This guide shows how to return custom errors from your Rust smart contract.
+
+Defining and deriving
+=====================
+
+Custom error codes help communicate why a contract rejects and can be returned
+both during initialization and during updates.
+
+On-chain, smart contracts return a numeric error code when rejecting. This is
+also the case, when using a custom error type. Therefore, a mapping from the
+custom error type to ``Reject``, in the form of an implementation of
+``From<MyError> for Reject``, is needed. We can also derive it
+automatically using ``#[derive(Reject)]``::
+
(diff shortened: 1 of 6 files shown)


source/smart-contracts/guides/custom-errors.rst
```diff
@@ -0,0 +1,55 @@
...
+This guide shows how to return custom errors from your Rust smart contract.
+
+Defining and deriving
+=====================
+
+Custom error codes help communicate why a contract rejects and can be returned
+both during initialization and during updates.
+
+On-chain, smart contracts return a numeric error code when rejecting. This is
+also the case, when using a custom error type. Therefore, a mapping from the
```
abizjak, line 16:
```suggestion
also the case when using a custom error type. Therefore, a mapping from the
```

source/smart-contracts/guides/custom-errors.rst
```diff
@@ -0,0 +1,55 @@
...
+
+   The valid range of error codes is ``i32::MIN..-1``. When deriving
+   ``Reject``, each variant is assigned an error code as determined by the
+   ordering. First variant (``ErrOne`` in example) gets ``-1``, second variant
+   (``ErrTwo`` in example) gets ``-2``, and so on.
+
+.. warning::
+
+   Deriving ``Reject`` is only possible for fieldless enums, i.e., enums where
+   the variants do not have associated data.
```
abizjak, line 37: Maybe also point out they should not have custom discriminants, e.g., they should not have things like
```rust
enum Foo {
   Foo = 3,
   Bar = 7,
}

abizjak (commented):
Apart from the two minor comments I think this is OK.

It could use more elaboration, but at least it is visible now, and can be improved in the future when we have more dedicated time.

abizjak (approved):

PR #104, opened 2021-05-18
Include extra files when building

limemloh:
## Purpose

Allow releasing files outside of sphinx

## Changes

The build script will now copy the `extra` directory into the build

I have remove the unused term and conditions as this will be an `extra` file instead


abizjak (approved):
Presumably the terms and conditions are included in the extra files in some other PR?

PR #107, opened 2021-05-24
Reorganized table of contents

### LINKED ISSUE #105: Restructure Table of Contents
**Task description**

Restructure content so that there's a clearer division between main categories such as Installation, Getting started, Concepts, Reference, and Resources and between Mobile Wallet, Desktop Wallet, and Concordium Client.

**Sub-tasks**
- [ ] Move existing topics into new categories (where needed).
- [ ]  Separate conceptual and task-based information in existing topics.

LotteKh:
## Purpose

Restructured table of contents for mainnet

## Changes

Added new category to ToC
Moved some files to new categories
Updated some links
Moved image folder

No content has been changed only the structure so there's no content review for this pull request.

## Checklist

- [ ] My code follows the style of this project.
- [x ] The code compiles without warnings.
- [x ] I have performed a self-review of the changes.
      hard-to-understand areas.

Closes #105


.vscode/settings.json
```diff
@@ -1,5 +1,5 @@
 {
-    "restructuredtext.confPath": "${workspaceFolder}/source",
+    "restructuredtext.confPath": "${workspaceFolder}\\source",
```
abizjak, line 2: I don't think this should be necessary. `/` is a valid path separator on all three platforms, but `\\` is only valid on Windows.

source/testnet/desktop-wallet/create-identity-desktop.rst
```diff
@@ -0,0 +1,57 @@
...
+
+#. Connect your Ledger hardware wallet if you haven't done so already and enter your pin code on the Ledger. Press the buttons above the up and down arrows to choose a digit, and then press both buttons to select the digit. When you've entered the PIN, press both buttons, and when it says **Concordium** press both buttons again. The Ledger says **Concordium is ready**.
+
+#. In the Desktop Wallet select **Submit**, when it says **Ledger Nano S is ready**. 
+
+#. Press both buttons to confirm the export of the following on the Ledger:
+
+   - Export PRF key
+   - Export IdCredSec
+   - Public-key
```
abizjak, line 44: Wouldn't it make sense to have some links to what `PRF key` and `IdCredSec` are? Otherwise it seems impossible for the user to understand that the first two are sensitive and secret, and the public key is not sensitive material.

source/testnet/desktop-wallet/create-identity-desktop.rst
```diff
@@ -0,0 +1,57 @@
...
+
+#. Press the right button to review the transaction on the Ledger. Verify that the following information on Ledger is identical to the information in the Desktop Wallet.
+
+   - Identity Credentials Public (IdCredSec)
```
abizjak, line 48: I am not sure what the Desktop wallet says, but `IdCredSec` has a dual `IdCredPub` and I would expect the latter to be `Identity Credentials Public`. Is this a typo?

source/testnet/desktop-wallet/create-identity-desktop.rst
```diff
@@ -0,0 +1,57 @@
...
+   - Registration ID (RegId)
+   - Verification Key
+   - Threshold
+
+#. The Ledger says **Sign transaction**. Press both buttons to sign the transaction.
+
+#. In the Desktop Wallet, enter the identity verification information and select **Submit**. This information will vary depending on the identity provider. The identity provider verifies your identity and submits the initial account to the blockchain. When the identity provider has verified your identity and submitted it to the blockchain, there's a green check mark next to the Concordium logo on the identity.
```
abizjak, line 55: In some other parts of the documentation there is a distinction between "submitted" and "confirmed/finalized". Here the identity provider submits it and waits until the initial account is created (that is, it is created in a finalized block), and then and only then confirms the identity. Thus the green checkmark should only appear at this point.

source/testnet/desktop-wallet/create-identity-desktop.rst
LotteKh: Yes, we need this information, but the content is not there yet.

PR #115, opened 2021-05-31
Run a node on Ubuntu and removal of troubleshooting

LotteKh:
## Purpose

Add topic on how to run a node with Ubuntu and removed troubleshooting topic and links to the topic. Also include a few very minor changes to other topics.


source/testnet/guides/run-node-ubuntu.rst
```diff
@@ -0,0 +1,110 @@
...
+This guide describes how organizations can run a node on the Concordium network from a server and how to set up the node to run as a :ref:`baker node<baker-node-Ubuntu>`.
+
+Prerequisites
+=============
+
+-  Ubuntu 20.04 must be installed on the server that is running the node.
+
+-  The server must be running around the clock.
+
+-  If you want to run the node as a baker, you must have generated baker keys.You can generate the keys in the Desktop Wallet. See :ref:`create-baker-desktop`.
```
abizjak, line 20:
```suggestion
-  If you want to run the node as a baker, you must have generated baker keys. You can generate the keys in the Desktop Wallet. See :ref:`create-baker-desktop`.
```

source/testnet/guides/run-node-ubuntu.rst
```diff
@@ -0,0 +1,110 @@
...
+
+.. Note::
+   If you want more detailed information about building and maintaining a node, or if your node is not running, see the `Building .deb packages for ubuntu distributions README from Concordium <https://github.com/Concordium/concordium-node/blob/main/scripts/distribution/ubuntu-packages/README.md>`__
+
+.. _baker-node-Ubuntu:
+
+Run a baker node on Ubuntu
+==========================
+
+The following steps show you how to run a node as a :ref:`baker <baker-concept>` on a server that participates in the Concordium network. When a node is a baker node, it receives blocks and transactions from other nodes and propagates information about blocks and transactions to the nodes in the Concordium network.
```
abizjak, line 53: This description is for any node, i.e., any node does this.

What a baker node does in addition to a passive node is that it participates in the lottery (and finalization if its stake is high enough) and produces its own blocks.

abizjak (approved):
I would rephrase the paragraph about what being a baker node means, but otherwise I think this is fine. 

We can refine and add some details later perhaps, if we see people are struggling.

PR #117, opened 2021-05-31
Hide testnet4 docs

limemloh:
## Purpose

Hide outdated testnet4 documentation

## Changes

- Redirect every page to an under-construction page.


limemloh:
This was suppose to be a MR for the testnet4 branch and not `main`, see #118 instead

PR #122, opened 2021-06-01
Add support and feedback section to be appended in every rst file

limemloh:
## Purpose

To ensure consistency, the support and feedback footer should be defined once and reused across the pages.

## Changes

- Added the footer using the sphinx setting [`rst_epilog`](https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-rst_epilog)
- Removed the support and feedback section from every rst file


LotteKh (approved):
Great!

PR #123, opened 2021-06-01
Disable the next and previous buttons in the bottom

limemloh:
## Purpose

Since the documentation is not structured as a book, with a start and a finish, it is better not to show the navigation pages in the bottom.

## Changes

- Remove next and previous buttons in the bottom of a page


LotteKh (approved):
Great - thanks

PR #168, opened 2021-06-16
Added time information to baker topics

LotteKh:
## Purpose

Updated baker topics with information about time before changes are applied

## Changes
Updated intro paragraphs in baker topics.

### DIFF
--- a/source/shared/net/desktop-wallet/change-baker-stake.rst
+++ b/source/shared/net/desktop-wallet/change-baker-stake.rst
@@ -9,9 +9,7 @@ Update the staked amount
    :backlinks: none
    :depth: 1
 
-You can change the :ref:`staked amount <concepts-baker-stake>` on a baker account at any time. However, the staked amount will not change right away. If you increase the stake the change will take effect when the current epoch has elapsed.
-
-If you decrease the stake, the change will take effect after 7 days (which corresponds to 168 epochs). After this, the amount you’ve decreased the stake with is returned to your disposable balance.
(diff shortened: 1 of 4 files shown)


source/shared/net/desktop-wallet/change-baker-stake.rst
```diff
@@ -9,9 +9,7 @@ Update the staked amount
    :backlinks: none
    :depth: 1
 
-You can change the :ref:`staked amount <concepts-baker-stake>` on a baker account at any time. However, the staked amount will not change right away. If you increase the stake the change will take effect when the current epoch has elapsed.
-
-If you decrease the stake, the change will take effect after 7 days (which corresponds to 168 epochs). After this, the amount you’ve decreased the stake with is returned to your disposable balance.
+You can change the :ref:`staked amount <concepts-baker-stake>` on a baker account at any time. If you increase the stake, the new stake will be applied after two :ref:`epochs <glossary-epoch>`, which corresponds to two hours. However, if you decrease the stake, there is a cool-down period before the new stake is applied. The cool down period is 7 days, which corresponds to 168 epochs. During this period, you'll not be able to remove the baker account or increase the stake. After the cool-down period, the amount you’ve decreased the stake with is returned to your disposable balance.
```
limemloh, line 12:
```suggestion
You can change the :ref:`staked amount <concepts-baker-stake>` on a baker account at any time. If you increase the stake, the new stake will be applied after two :ref:`epochs <glossary-epoch>`, which corresponds to two hours. However, if you decrease the stake, there is a cool-down period before the new stake is applied. The cool down period is 7 days, which corresponds to 168 epochs. During this period, you'll not be able to remove the baker account or further update the stake. After the cool-down period, the amount you’ve decreased the stake with is returned to your disposable balance.
```
Since you are also not able to decrease it further.

limemloh, line 12: Also I think the first sentence is a bit misleading, because you can not change the stake during a cooldown.
If you make transactions for changing the stake while in cooldown, the transaction will be rejected in the block.

limemloh (approved):
Some minor comments, otherwise looks good to me

PR #174, opened 2021-06-17
Update testnet downloads

LotteKh:
## Add testnet links to Testnet downloads page

Remove mainnet links from testnet downloads

## Changes

-deleted shared downloads file
- added separate downloads pages for testnet and mainnet
- changed links for testnet and removed links to mobile wallet, desktop wallet, and ledger app from testnet downloads.


source/mainnet/net/installation/downloads.rst
```diff
@@ -1 +1,118 @@
-.. include:: /shared/net/installation/downloads.rst
+
+.. _downloads:
```
limemloh, line 2: This label should be moved to above the first header in this file. 
Otherwise `variables.rst` could potentially change link of the label

source/testnet/net/installation/downloads.rst
```diff
@@ -1 +1,89 @@
-.. include:: /shared/net/installation/downloads.rst
+.. _downloads:
```
limemloh, line 1: Same as the comment for mainnet

limemloh (changes requested):

limemloh (approved):

PR #183, opened 2021-06-23
Ledger app installation edits

LotteKh:
## Purpose
Describe that "broken certificate chain" should be ignored when installing the certificate.

## Changes

Edited certificate installation description
Indented code examples


Bargsteen (approved):
Looks good!

PR #221, opened 2021-08-31
Various minor changes

LotteKh:
## Purpose

Fix minor issues

## Changes

Minor edit such as missing bold UI elements in a number of files.

### DIFF
--- a/source/shared/net/desktop-wallet/update-baker-keys.rst
+++ b/source/shared/net/desktop-wallet/update-baker-keys.rst
@@ -90,7 +90,7 @@ When the co-signers have signed the transaction, they return the signed transact
 
 #. Select the transaction that you want to submit to the chain. You can see an overview of the transaction details and an overview of the signatures. You can also see that the status of the transaction is Unsubmitted, and you can see the identicon and the transaction hash.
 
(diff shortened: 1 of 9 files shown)


source/shared/net/desktop-wallet/update-baker-keys.rst
```diff
@@ -101,9 +101,9 @@ When you have received and added all the required signatures, you can submit the
 
 #. Review the transaction details carefully to ensure that all information is correct.
 
-#. Select I understand this is the final submission and that it cannot be reverted.
+#. Select **I understand this is the final submission and that it cannot be reverted**.
 
-   If you don’t want to submit the transaction to the chain, you can select **Cancel**. The proposal is no longer active. However, it is still visible in the list of proposals.
+   If you don’t want to submit the transaction to the chain, you can select **Cancel**. The proposal is no longer active but still visible in the list of proposals.
```
jens-concordium, line 106:
```suggestion
   If you don’t want to submit the transaction to the chain, you can select **Cancel**. The proposal is no longer active, but it's still visible in the list of proposals.
```
I'm not sure which option is best, so feel free to discard my suggestion

jens-concordium (approved):
Just had the one comment, which might not even be necessary to change

PR #248, opened 2021-09-22
Update Ledger app

LotteKh:
## Purpose

Describe how to update the Ledger app to the latest version.

## Changes

Added sections on how to update the Ledger app for Windows, macOS, and Ubuntu.

### DIFF
--- a/source/shared/net/desktop-wallet/install-ledger-app.rst
+++ b/source/shared/net/desktop-wallet/install-ledger-app.rst
@@ -108,7 +108,7 @@ You now have to install a custom certificate on the Ledger to ensure that it tru
 
 #. Enter your PIN.
 
-.. _install-app-windows:
+.. _install-ledger-app-windows:
 
 Install the Concordium Ledger app on Windows
 --------------------------------------------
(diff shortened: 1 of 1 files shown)


source/shared/net/desktop-wallet/install-ledger-app.rst
```diff
@@ -108,6 +108,8 @@ You now have to install a custom certificate on the Ledger to ensure that it tru
 
 #. Enter your PIN.
 
+.. _install-app-windows:
```
jens-concordium, line 111:
```suggestion
.. _install-ledger-app-windows:
```
Might be good to add "ledger" to the link, in case we get other Windows apps at some point.

source/shared/net/desktop-wallet/install-ledger-app.rst
```diff
@@ -184,6 +196,8 @@ You now have to install a custom certificate to ensure that the Ledger trusts ap
 
 #. Enter your PIN.
 
+.. _install-app-macos:
```
jens-concordium, line 199:
```suggestion
.. _install-ledger-app-macos:
```
Same comment as above :)

source/shared/net/desktop-wallet/install-ledger-app.rst
```diff
@@ -261,6 +284,8 @@ You now have to install a custom certificate to ensure that the Ledger trusts ap
 
 #. The certificate is installed on the Ledger. Press the right button to navigate through the key, and then press both buttons when the Ledger says **Trust certificate**.
 
+.. _install-app-ubuntu:
```
jens-concordium, line 287:
```suggestion
.. _install-ledger-app-ubuntu:
```
And the same here as well

jens-concordium (approved):
Just had the three small comments about adding "-ledger-" to the links. Maybe it would also be good to tell users to make sure they have the right version of the Ledger firmware, before updating? Right now it's not necessary I guess, but it might come in handy later.

PR #261, opened 2021-09-28
Ledger terminology and memo changes

LotteKh:
## Purpose

Update topics with new Ledger terminology and add information about adding memos to transactions.

## Changes

Updated procedures (both single and multi-signature transactions) with new terminology and memo description:
Simple transfers
Shielded transfers
Encrypted transfers

### DIFF
--- a/source/shared/net/desktop-wallet/send-shielded-amount.rst
+++ b/source/shared/net/desktop-wallet/send-shielded-amount.rst
@@ -1,7 +1,7 @@
 .. _send-shielded-amount:
 
 ====================================
-Send and receive a shielded transfer
+Send and decrypt a shielded transfer
 ====================================
 
 .. contents::
@@ -73,6 +73,7 @@ Any shielded transfers that you've already decrypted are visible in the shielded
 #. In the Desktop Wallet, select **Decrypt**.
 
 #. On the Ledger, press both buttons to decrypt the amount. In the Desktop Wallet you can now see the amount that you decrypted.
+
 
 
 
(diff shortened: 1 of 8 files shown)


source/shared/net/desktop-wallet/send-shielded-amount.rst
```diff
@@ -1,8 +1,8 @@
 .. _send-shielded-amount:
 
-========================
-Make a shielded transfer
-========================
+====================================
+Send and receive a shielded transfer
```
jens-concordium, line 4: Depending on whether or not the subtitle below is changed or not, this might have to be updated as well.

source/shared/net/guides/multisig-transfer.rst
```diff
@@ -62,7 +62,14 @@ Make a transfer proposal
...
+#.  In **Amount**, enter the amount of GTU you want to transfer. If the amount you want to transfer including the transfer fee exceeds the amount in your account, you’ll see a message saying **Insufficient funds**. This means that there's not enough GTU in the account and you’ll have to change the amount. (The amount must also be greater than zero.) In the **Transaction details** under **Amount**, you can see the estimated fee.
+
+#. Optionally, in **You can enter a memo here**, enter a **Memo** if you want to send a message to the recipient along with the transfer. Note that there's a fee associated with sending a memo and the fee depends on the size of the memo in bytes.
+
+   .. Warning::
+      Transaction memos are public and visible on the blockchain, and they are permanent and can't be deleted. Therefore, carefully review all information before you submit the memo to the blockchain.
```
jens-concordium, line 70:
```suggestion
     Transaction memos are public, permanent and visible to everyone on the blockchain. Therefore, carefully review all information before you submit the memo to the blockchain.
```
If this is changed in the other file, it should also be changed here.

jens-concordium (approved):

PR #268, opened 2021-09-30
Run baker node on windows

LotteKh:
## Purpose

Move windows and macOS node topics to mainnet. Update information and update cross-references in baker andnode

## Changes

-Moved Windows node topic
-Updated Windows node topic with mainnet info
-Moved macOS topic
-Updated xrefs in several files


source/shared/net/guides/baker-windows.rst
```diff
@@ -0,0 +1,58 @@
...
+   .. Note::
+
+      The baker credentials file must be accessible by the SYSTEM user and can’t be encrypted.
+
+   You can find the node logs here:
+
+   - Mainnet: ``C:\ProgramData\Concordium\Node Runner\mainnet\logs``
+   - Testnet: ``C:\ProgramData\Concordium\Node Runner\testnet\logs``
+
+   For more information on the logs, see :ref:`Run a node on Windows<run-node-windows>`.
```
td202, line 53: Perhaps link directly to the logs section of that page.

We could possibly add a troubleshooting section with the following scenarios:

- The node does not restart after setting the credentials. This is likely to be due the location of the baker credentials file being incorrectly specified. This can be confirmed by an event such as the following in the application log in the Windows Event Viewer:
image
In this case, correct the path in the configuration file.

- The node does not start baking and "Baker keys are incorrect" appears (repeatedly) in the log file. It takes 1-2 hours after baker keys have been added to an account for the keys to be eligible for baking. If this problem persists beyond two hours, then the keys likely are incorrect. Double check that you are not using keys from testnet on mainnet, or vice-versa.

source/shared/net/guides/run-node-windows.rst
```diff
@@ -0,0 +1,213 @@
...
+
+   - **Automatically start nodes at system startup**:  When selected, the node runs when the system starts. Choose this option when you plan to use the node frequently and need it to be up-to-date at short notice. If you don’t select this option, you’ll have to start the node manually when required, for example, when you want to use the Desktop Wallet or concordium-client. If you choose to start the node manually, it might take longer for the node to get up-to-date with the blockchain depending on when the node was last up-to-date.
+
+   - **Start nodes after installation is complete**: The node will automatically start running after the installation is complete. Do not choose this option if you want to make further configuration changes before starting the node.
+
+   - **Mainnet** or **Testnet**: Mainnet is the Concordium production blockchain where actual transactions occur, whereas testnet is for test and development. Select your preferred network for running the node: **Run a mainnet node** or **Run a testnet node**. You can run nodes both on the mainnet and on the testnet, and you can run nodes on both networks at the same time.
```
td202, line 46:
```suggestion
   - **Mainnet** or **Testnet**: Mainnet is the Concordium production blockchain where actual transactions occur, whereas testnet is for test and development. Select your preferred network for running the node: **Run a mainnet node** or **Run a testnet node**. You can run a node on just the mainnet or the testnet, or run nodes on both at the same time.
```

source/shared/net/guides/run-node-windows.rst
```diff
@@ -0,0 +1,213 @@
...
+     .. image:: ../images/run-node/Node-setup-win-3.png
+         :width: 50%
+
+Verify that the node is running
+===============================
+
+The node runs as a background service with no user interface. To verify that it’s running, use the **Task Manager**.
+
+#. Search for *task* in the **Search** bar, and then select **Task Manager**.
+#. Select the **Details** tab, and verify that the following is running:
```
td202, line 77:
```suggestion
#. Select the **Details** tab, and verify that the following are running:
```

source/shared/net/guides/run-node-windows.rst
```diff
@@ -0,0 +1,213 @@
...
+
+
+Connect a node to the Desktop Wallet
+====================================
+
+You can also verify that a node is running by connecting it to the Desktop Wallet.
+
+#. In the Desktop Wallet, go to **Settings**, and then select **Node settings**.
+
+   - If you're running the mainnet version of the Desktop Wallet, you must connect to a mainnet node. In the **Address field**, enter *127.0.01* and in the **Port field** enter *10000*.
```
td202, line 105:
```suggestion
   - If you're running the mainnet version of the Desktop Wallet, you must connect to a mainnet node. In the **Address field**, enter *127.0.0.1* and in the **Port field** enter *10000*.
```

td202 (changes requested):
A couple of suggestions and a few minor things. Otherwise, looks good!

jens-concordium (commented):
I didn't have a lot of comments. One thing I noticed, was that we seem to be a little inconsistent in whether we write Mainnet and Testnet with capitalized letters or not. I'm not sure if we ever made a decision on that, but maybe we should?

source/shared/net/guides/run-node-windows.rst
```diff
@@ -0,0 +1,213 @@
...
+Synchronize a node with the network
+===================================
+
+When you start a node for the first time, it can take a while to synchronize the node with the rest of the network, since it has to get all blocks from its peers.
+
+You can improve the performance by downloading the blocks before starting the node. While it will still take time to process the blocks, it will typically be faster than requesting them from peers.
+
+#. Download the file with the blocks from the following addresses:
+
+   - Mainnet: https://dashboard.mainnet.concordium.software/blocks_to_import.mdb
```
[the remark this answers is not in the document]
td202: Should be: https://catchup.mainnet.concordium.software/blocks_to_import.mdb

jens-concordium (approved):
I just had two comments, which might not even need to be attended.

source/shared/net/guides/baker-windows.rst
```diff
@@ -21,24 +21,20 @@ Prerequisites
...
 
-   You register the baker credentials either on the mainnet or on the testnet. We recommend that you add the baker credentials for the network that you created the credentials on. Don’t use the same baker credentials on both networks.
-
-   -  If you created the credentials in a mainnet version of the Desktop Wallet, add the credentials to ``node.mainnet``.
-
-   -  If you created credentials in a testnet version of the Desktop Wallet, add the credentials to ``node.testnet``.
-
-#. Depending on whether you're running a node on testnet or mainnet or both, add the following in the ``[node.mainnet]`` or ``[node.testnet]`` section:
+#. Add the following in the [node.mainnet] section if the credentials are for ``[node.mainnet]`` or ``[node.testnet]`` section if they are for testnet. Make sure that you add the credentials to the network the account was created on.
```
td202, line 28:
```suggestion
#. Add the following in the ``[node.mainnet]`` section if the credentials are for mainnet, or the ``[node.testnet]`` section if they are for testnet. Make sure that you add the credentials to the network the account was created on.
```

source/shared/net/guides/run-node-windows.rst
```diff
@@ -159,7 +161,17 @@ When nodes are stopped or started, this is also recorded in the system event log
...
+--------------------
+
+If a node fails to start, for example because of a configuration issue, or stops unexpectedly, there will be an Error-level event in the **Event Viewer** that explains why. The following are common error scenarios and possible solutions.
+
+- The nodes doesn't restart after setting the credentials. This error can occur when the specified location of the credentials file is incorrect. In the **Event Viewer** an error such as the following might appear. When this happens, change the path in the configuration file.
+
+   .. image:: ../images/run-node/Node-setup-win-10.png
+         :width: 60%
+
+- The node doesn't start baking, and in the log file you see the message *Baker keys are incorrect* repeatedly. It takes 1-2 hours after baker keys have been added to an account for the keys to be eligible for baking. If this problem persists beyond two hours, then the keys likely are incorrect. Double check that you are not using keys from testnet on mainnet, or vice-versa.
```
td202, line 174: These pertain specifically to when you set the baker credentials, so it may be better to put them in the guide on becoming a baker. Otherwise, it should be clarified here that these scenarios pertain to the situation where the user has set baker credentials in their configuration file. (In particular, in the context of this page, it is probably not clear what "setting the credentials" means> Also, if they haven't configured the baker then the node isn't going to start baking, but the second solution might be confusing without further context.)

td202 (changes requested):

source/shared/net/guides/baker-windows.rst
```diff
@@ -21,24 +21,20 @@ Prerequisites
 Configure a baker node on Windows
 =================================
 
-#. To run a Windows node as baker, you have to select an account as baker account in the Desktop Wallet and generate baker keys for the account. For more information, see :ref:`Add a baker account in the Desktop Wallet<create-baker-desktop>`. You get the account information by exporting a JSON file with the baker credentials from the Desktop Wallet.
+#. Create baker keys and associate them with an account either on mainnet or testnet. We recommend that you do this with the :ref:`Desktop Wallet<create-baker-desktop>`. However, if you want use an account from the mobile wallet as a baker account, you have to use the :ref:`Concordium Client<become-a-baker>`. In the following, the baker keys are referred to as baker-credentials.json.
```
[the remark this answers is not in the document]
LotteKh: Mainnet and testnet are terms used for blockchains in general. So since they are not proper nouns, I'd suggest that we don't capitalize. However, if we're going to use a name such as Concordium Mainnet, we should capitalize.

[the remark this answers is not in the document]
jens-concordium: That's a good point! Let's go with that.

td202 (approved):

PR #307, opened 2021-11-05
Changes including backup info to wallet comparison

LotteKh:
## Changes

Changed the topic Deciding between the Desktop Wallet and the Mobile Wallet

- Added more information and more links about backups
- Added more information about storing private keys
- Reorganized and edited the content of the topic


source/shared/net/guides/deciding-wallet.rst
```diff
@@ -1,79 +1,100 @@
...
+The following provides more details on the differences between the wallets.
+
+How the wallets store your private keys
+=======================================
+
+One of the main differences between the two wallets lies in how they store your private keys.
+
+-  The Desktop Wallet stores your private keys on a Ledger that is secured by a PIN code and backed up by a recovery phrase. The private keys never leave the Ledger hardware device, and the Ledger doesn't require internet access, which means you can keep your private keys offline. However, this also means that you need the Ledger to sign the transactions you create in the Desktop Wallet, which makes it less convenient than the Mobile Wallet.
+
+- The Mobile Wallet creates and stores your private keys on the phone. This means it's easier to use because you don't need a Ledger device to make transactions. However, because your phone is connected to the internet, the wallet is more vulnerable to security breaches than the Desktop Wallet. Also, anyone with access to the mobile wallet can make transfers from the accounts in the mobile wallet.
```
smh1001, line 43: This is wrong: "Also, anyone with access to the mobile wallet can make transfers from the accounts in the mobile wallet."

The wallet is protected by passcode or fingerprint/faceid so only if the user hands out his passcode do other have that level of access. I think we should remove this sentence completely.

source/shared/net/guides/deciding-wallet.rst
```diff
@@ -1,79 +1,100 @@
...
+We strongly recommend that you make a backup of your wallet regardless of which one you use. Backups are created in different ways in the two wallets.
 
-Why do I need a Ledger hardware device to use the Desktop Wallet?
-==================================================================
+- **Backup of the Desktop Wallet**: You create a backup of your accounts, identities, and addresses by exporting them to a file from the Desktop Wallet. This doesn't mean that you create a backup of your private keys. The backup of your private keys is essentially the 24-recovery phrase for the Ledger. If you lose the PIN code to the Ledger, you can restore the Ledger device from your recovery phrase. You can also set up a new Ledger device with the recovery phrase. It's vital that you keep the recovery phrase safe. For more information, see :ref:`Make a backup of identities, accounts, and addresses<export-import-desktop>` and :ref:`Account recovery<account-recovery-desktop>`.
```
smh1001, line 49: Something missing here: "24-recovery"
Maybe 24-word recovery phrase or just seed phrase as Ledger calls it.

source/shared/net/guides/deciding-wallet.rst
```diff
@@ -1,79 +1,100 @@
...
-     - Yes
-   * - **Create and manage baker accounts**
-     - Yes
-     - The Concordium Client can be used to manage baker accounts. Rewards are visible in the wallet.
-   * - **Run a node**
-     - Requirement
-     - Not needed
+   *  - **Desktop Wallet**
+      - **Mobile Wallet**
+   *  - Secured by password
```
smh1001, line 74: Keys secured by Ledger device

source/shared/net/guides/deciding-wallet.rst
```diff
@@ -1,79 +1,100 @@
...
+      - Secured by password and biometrics
+   *  - Encrypted
+      - Encrypted
+   *  - Backup includes accounts, identities, and the address book.
```
smh1001, line 78: Is this right. Backup does not include account keys, right? Will people be able to distiguish between accounts and account private keys? So the main thing is probably identities

source/shared/net/guides/deciding-wallet.rst
```diff
@@ -1,79 +1,100 @@
...
+      - Backup includes accounts, identities, address book, and private keys.
+   *  - Private keys are stored on the Ledger that is secured by a PIN code and backed up by recovery phrase.
+      - Private keys are stored in the wallet.
+   *  - Creation and management of single signature accounts
+      - Creation and management of single signature accounts
+   *  - Creation and management of multi signature accounts
+      - No creation and management of multi signature accounts
```
smh1001, line 85: I think maybe this should just be NA. It is easy to miss the No in the beginning.

smh1001 (changes requested):

source/shared/net/guides/deciding-wallet.rst
LotteKh: Ledger uses both terms, but I think they use recovery phrase more than seed phrase.

source/shared/net/guides/deciding-wallet.rst
LotteKh: This item is mentioned further down in the list

source/shared/net/guides/deciding-wallet.rst
jens-concordium: I misread this the first time. Here's my new take after re-reading it:

What Stefan said about the identities being the main thing is right I think, as they can't be restored using only the Ledger, although I guess the same goes for the address book. Accounts technically can be restored from the Ledger and recovery phrase alone, but that doesn't contain the given account names, so there's sort of "two parts" to backing them up.

So maybe we should phrase it like the export contains names and addresses for the accounts, the identities and the address book. In combination with the Ledger, or a recovery phrase and a Ledger, the export can used to recover everything from the the desktop wallet. 

I don't know if it makes more sense now, @LotteKh?

jens-concordium (approved):
Looks good!

smh1001 (approved):

PR #322, opened 2021-11-24
Rename concordium-client transaction commands

LotteKh:
## Purpose

_Rename transaction commands after name change from GTU to CCD

## Changes

Changed commands in transactions.rst


abizjak (approved):
I think this covers the changes that were made now.

I can see there are still some remaining uses of --encrypted in concordium-client, but we can resolve those after the client has been updated as well.

PR #325, opened 2021-11-25
client transactions only changed for testnet not mainnet

LotteKh:
## Purpose

_Changes to concordium-client were included in both testnet and mainnet docs, but are currently only relevant for testnet.

## Changes

_Changed the topic Concordium Client transactions in testnet to include changes. Same topic for mainnet doesn't include changes.

### DIFF
--- a/source/shared/net/references/transactions.rst
+++ b/source/shared/net/references/transactions.rst
@@ -19,13 +19,13 @@ Transaction commands
 +-------------------------------+-------------------------------------+
 | Command                       | Description                         |
 +===============================+=====================================+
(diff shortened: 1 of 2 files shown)


jens-concordium:
In the code block on line 181 of the Mainnet (shared) documentation it says `send-shielded`, which should still be `send-gtu-encrypted`, I guess. 

The rest looks good.

jens-concordium (commented):
In the code block on line 181 of the Mainnet (shared) documentation it says `send-shielded`, which should still be `send-gtu-encrypted`.

The rest looks good.

jens-concordium (approved):

PR #329, opened 2021-12-01
Rename remaining uses of encrypted.

abizjak:
## Purpose

Some further renamings were done in concordium-client that are not reflected in the documentation at present.

## Changes

Rename concordium-client commands to reflect the new reality.


LotteKh (approved):

jens-concordium:
Note:
Some of these changes are made in the "shared" folder, which means they will go out for both Mainnet and Testnet when deployed. So far, we have only deployed the name change related updates for the Testnet specific documentation, as the updated client has only been released for that.

Let's wait with updating the Mainnet docs until the release later in december, and then make the changes of this PR in the Testnet specific documentation. 

@LotteKh and I will look into it.

PR #373, opened 2021-12-22
Added recommendations from moderator in issue 633.

dg-concordium:
## Purpose

To clarify what commands users running Concordium client should do and to add a note that the baker credentials file should not be changed.

## Changes

Added information to step one in the configure section.


dg-concordium:
This also needs the Edit in GitHub button removed. https://github.com/sphinx-doc/sphinx/issues/2386

jens-concordium (approved):

PR #378, opened 2022-01-04
Correct finalizer requirement

Bargsteen:
## Purpose

Correct a mistake about the requirement for becoming a finalizer.

## Changes

Correct the phrase, so it matches reality.
Used same phrasing as seen in https://developer.concordium.software/en/mainnet/net/guides/become-baker.html#finalization

## Checklist

      hard-to-understand areas.

### DIFF
--- a/source/shared/net/resources/glossary.rst
+++ b/source/shared/net/resources/glossary.rst
@@ -180,7 +180,7 @@ Finalization
 ============
 
 The process by which a block is marked to be "finalized", i.e. part of the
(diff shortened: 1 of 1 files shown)


abizjak (approved):

dg-concordium (approved):

dg-concordium:
I was working on text for the same (for the Baker conceptual topic) and had come across it written in two different ways. So this answered my question.

PR #381, opened 2022-01-07
Added Enabling inbound connections text to topics.

dg-concordium:
## Purpose

To clarify for users about opening up port 8888 on firewall to enable more connections.

## Changes

Added text from Docker topic to Windows, macOS, and Ubuntu topics.

## Checklist

- [X ] My code follows the style of this project.
- [X ] The code compiles without warnings.
- [X ] I have performed a self-review of the changes.
      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

- [ X] I accept the above linked CLA.

### DIFF
--- a/source/mainnet/net/guides/run-node-ubuntu.rst
+++ b/source/mainnet/net/guides/run-node-ubuntu.rst
@@ -86,9 +86,9 @@ Concordium network. It will be able to send transactions and,
 
 However you can also make your node an even better network participant
 by enabling inbound connections. By default, ``concordium-node`` listens
-on port ``8888`` for inbound connections. Depending on your network and
+on port ``8888`` for inbound connections on Mainnet and on port ``8889`` for inbount connections on Testnet. Depending on your network and
 platform configuration you will either need to forward an external port
-to ``8888`` on your router, open it in your firewall, or both. The
+to ``8888`` or ``8889`` on your router, open it in your firewall, or both. The
 details of how this is done will depend on your configuration.
 
 .. _baker-node-Ubuntu:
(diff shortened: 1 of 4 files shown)


source/mainnet/net/guides/run-node-ubuntu.rst
```diff
@@ -74,6 +74,23 @@ The ``concordium-mainnet-node`` service that you just installed will be running
...
+If you are running your node behind a firewall, or behind your home
+router, then you will probably only be able to connect to other nodes,
+but other nodes will not be able to initiate connections to your node.
+This is perfectly fine, and your node will fully participate in the
+Concordium network. It will be able to send transactions and,
+:ref:`if so configured<become-a-baker>`, to bake and finalize.
+
+However you can also make your node an even better network participant
+by enabling inbound connections. By default, ``concordium-node`` listens
+on port ``8888`` for inbound connections. Depending on your network and
```
abizjak, line 89: The ubuntu mainnet node by default listens on 8888, but the testnet node by default listens on 8889. 

I believe the same applies to Windows and Mac nodes. The ports have to be different so both can run at the same time.

abizjak (changes requested):

source/mainnet/net/guides/run-node-ubuntu.rst
dg-concordium: I need confirmation on the Windows and Mac. Who should I ask? Or when you say "I believe" is it not just a belief but a fact?

abizjak: I checked again, and it is indeed as I wrote also for WIn & Mac.

smh1001 (approved):

abizjak (approved):
Looks good.

I think at some point we should probably revise the first paragraph to make it more likely for people to do this. But it's OK as it is for now.

PR #392, opened 2022-01-12
Updated note about transaction fees and updated/added to many topics …

dg-concordium:
## Purpose

To clarify that the user needs to be sure that they have sufficient funds to cover transactions.

## Changes

Updated the transaction fees note and then updated it or added it to many topics for DW, MW, Concordium Client and the Transactions reference topic.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


td202 (commented):
I think the note is fine, but it buries the lede a bit. Particularly in the context of shielding, I'd like a big red warning to the effect: Don't shield all of your funds as you won't be able to transfer or unshield them (without additional funds)!

Then they can read the note to get a better understanding of why they shouldn't, and how much they should hold back.

dg-concordium:
> I think the note is fine, but it buries the lede a bit. Particularly in the context of shielding, I'd like a big red warning to the effect: Don't shield all of your funds as you won't be able to transfer or unshield them (without additional funds)!
> 
> Then they can read the note to get a better understanding of why they shouldn't, and how much they should hold back.

I've added a stronger warning at the top of the Shield - unshield topic for Mobile Wallet. Should I add a similar note for some of the other transactions where users can get themselves into this mess?

Could there be some sort of warning that pops up in the app when the user tries to shield more than a certain percentage of the balance? I mean, we can't stop them from doing this, but we could at least warn them.

td202 (approved):

PR #398, opened 2022-01-28
Added note about default path in run-node-ubuntu

chrmatt:
People may wonder whether they need to adjust the given path to something related to their setup. This is not the case. This update clarifies that the string is the genesis hash.

Furthermore, the testnet documentation used the same hash as the mainnet documentation. This was a mistake since testnet has a different genesis hash.

## Purpose

This pull request clarifies the documentation and fixes a bug in the testnet documentation.

## Changes

Added a note explaining the meaning of the provided path and fixed hash in testnet documentation.

## Checklist

      hard-to-understand areas.

### DIFF
--- a/source/mainnet/net/guides/run-node-ubuntu.rst
+++ b/source/mainnet/net/guides/run-node-ubuntu.rst
@@ -120,7 +120,7 @@ Configure the node with baker keys
       BindReadOnlyPaths=/home/user/concordium/baker-credentials.json:%S/concordium-9dd9ca4d19e9393877d2c44b70f89acbfc0883c2243e5eeaecc0d1cd0503f478/baker-credentials.json
 
    Where you replace the path `/home/user/concordium/baker-credentials.json` with the actual location of the file.
-   
+
 .. Note::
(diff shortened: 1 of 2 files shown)


source/mainnet/net/guides/run-node-ubuntu.rst
```diff
@@ -121,6 +121,9 @@ Configure the node with baker keys
 
    Where you replace the path `/home/user/concordium/baker-credentials.json` with the actual location of the file.
 
+.. Note::
+   The path `%S/concordium-9dd9ca4d19e9393877d2c44b70f89acbfc0883c2243e5eeaecc0d1cd0503f478/` is the default path to the baker's state directory, where `9dd9ca4d19e9393877d2c44b70f89acbfc0883c2243e5eeaecc0d1cd0503f478` is the genesis hash.
```
abizjak, line 125: Could you also add this note to the testnet version of the documentation (changing the hash appropriately of course).

chrmatt: It actually seems the same hash is used in the testnet documentation!? Is this a bug in the testnet document or did I mix up the documents?

abizjak: This is a bug in the documentation, the testnet genesis is, of course, 
`b6078154d6717e909ce0da4a45a25151b592824f31624b755900a74429e3073d`

So this PR is two for one :)

abizjak (approved):

source/mainnet/net/guides/run-node-ubuntu.rst
chrmatt: Should be fixed.

PR #417, opened 2022-02-18
Added copy button function to code blocks.

dg-concordium:
## Purpose

To improve the user experience, add a copy button to code blocks to ensure that the entire thing is copied.

## Changes

Updated conf.py for mainnet and testnet to add extension for copy button. Added another extension as well but there is a known bug in this so it is commented out.

The copy button should be checked to ensure that it copies what is expected as expected. No configuration of the button was done but it does have options. [https://sphinx-copybutton.readthedocs.io/en/latest/]

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/source/mainnet/conf.py
+++ b/source/mainnet/conf.py
@@ -23,7 +23,7 @@ sys.path.append(os.path.abspath('../exts'))
 # -- Project information -----------------------------------------------------
 
 project = 'Concordium'
-copyright = '2021, Concordium Software ApS'
+copyright = '2021 - 2022, Concordium Software ApS'
 author = 'Concordium'
 
 # The short X.Y version
--- a/source/mainnet/conf.py
+++ b/source/mainnet/conf.py
@@ -49,7 +49,9 @@ extensions = [
     "sphinx.ext.extlinks",
     # "sphinx.ext.imgconverter", # To support svg when targeting LaTeX
     "multidoc",
-    "sphinx_reredirects"
+    "sphinx_reredirects",
+    # "sphinx_rtd_dark_mode", # A bug in this prevents code blocks from displaying correctly in dark mode
+    "sphinx_copybutton"
 ]
 
 # sphinx-prompt must be the first of these two.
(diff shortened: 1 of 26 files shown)


jens-concordium (approved):

source/mainnet/conf.py
```diff
@@ -49,7 +49,9 @@
     "sphinx.ext.extlinks",
     # "sphinx.ext.imgconverter", # To support svg when targeting LaTeX
     "multidoc",
-    "sphinx_reredirects"
+    "sphinx_reredirects",
+    # "sphinx_rtd_dark_mode", # A bug in this prevents code blocks from displaying correctly in dark mode
+    "sphinx_copybutton"
 ]
```
abizjak, line 56: I would recommend adding

```
copybutton_prompt_text = "$"
copybutton_line_continuation_character = "\\"
```
here (and the same for testnet). This will mean that, for example, in the following 
Screenshot from 2022-02-26 20-40-02

only the text
```
concordium-client config --help
```
is copied instead of both the input and output.

At least I think this is the most reasonable behaviour, don't you think?

source/testnet/conf.py
```diff
@@ -48,7 +48,9 @@
     "sphinx.ext.intersphinx",
     # "sphinx.ext.imgconverter", # To support svg when targeting LaTeX
     "multidoc",
-    "sphinx_reredirects"
+    "sphinx_reredirects",
+    # "sphinx_rtd_dark_mode", # A bug in this prevents code blocks from displaying correctly in dark mode
+    "sphinx_copybutton"
 ]
```
abizjak, line 54: The same thing as for mainnet should be added here.

abizjak (commented):
I think this'll look good (provided the two options I mentioned are added), but first https://github.com/Concordium/concordium.github.io/pull/423/files should be merged into this branch so that copying and stripping of the `$` will work for all examples.

source/mainnet/conf.py
dg-concordium: Yes. Have added both lines.

abizjak (approved):

PR #514, opened 2022-06-13
Fix schema json topic

### LINKED ISSUE #512: Schema reference for enums split in two
**Bug Description**

The reference for enum schemas seems to be split in two. There is the part under the header (with `Some`), and then there is a part at the bottom with `None`. They should be grouped together, I believe.

image

So `Example of JSON...` in this image should be moved up to the enum section.
image

dg-concordium:
## Purpose

An example had been split, causing confusion for readers.

## Changes

Merge empty into enum section to make example clearer for users. Added cross reference to enum in Empty.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

Closes #512

### DIFF
--- a/source/mainnet/smart-contracts/references/schema-json.rst
+++ b/source/mainnet/smart-contracts/references/schema-json.rst
@@ -207,26 +207,18 @@ Example of JSON for an enum ``Option``:
        Some(U32)
    }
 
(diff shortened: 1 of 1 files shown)


Bargsteen (approved):
I changed it a bit. But it looks good now 😊

PR #539, opened 2022-06-27
Inline instructions for out of band catchup on ubuntu.

abizjak:
## Purpose

Addresses #https://github.com/Concordium/concordium-node/issues/367

## Changes

Add instructions directly to the documentation instead of linking to the repository readme.


dg-concordium (approved):

abizjak:
Thank you @dg-concordium :)

PR #611, opened 2022-09-23
Update downloads with command to run in Linux

dg-concordium:
## Purpose

Inform users they need to run a command when they have downloaded the Linux version of Concordium Client to make the file executable.

## Changes

Added to downloads pages. On the testnet page I cannot remove the extra line space because Sphinx does not like it if there is not a blank line after the bullet. And because it is variable text and not regular text I have to put in a pipe to force space. Otherwise the text for the addition is jammed up against the checksum.

Also minor corrections to concordium client topic.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/source/mainnet/net/installation/downloads-testnet.rst
+++ b/source/mainnet/net/installation/downloads-testnet.rst
@@ -91,7 +91,9 @@ Concordium Client v4.1.0
 
       .. code-block:: console
 
-         chmod +x concordium-client
+         chmod +x concordium-client_*
+
+   where you replace `*` with |client-version|.
 
 -  `Download the Testnet Concordium Client for macOS <https://distribution.concordium.software/tools/macos/signed/concordium-client-4.1.0.pkg>`_
 
--- a/source/mainnet/net/installation/downloads-testnet.rst
+++ b/source/mainnet/net/installation/downloads-testnet.rst
@@ -86,7 +86,7 @@ Concordium Client v4.1.0
    - SHA256 checksum of the download: |client-linux-checksum|
 
    |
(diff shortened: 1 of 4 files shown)


source/mainnet/net/installation/downloads-testnet.rst
```diff
@@ -85,6 +85,14 @@ Concordium Client v4.1.0
 
    - SHA256 checksum of the download: |client-linux-checksum|
 
+   |
+
+   When you run Concordium Client on Linux, run the following command to make the file you download executable:
+
+      .. code-block:: console
+
+         chmod +x concordium-client
```
abizjak, line 94: The file that they download will be called (at the moment) `concordium-client_4.1.0` so technically they should run
```
chmod +x concordium-client_4.1.0
```

But it is not a good idea to write the exact version number since it'll be hard to maintain. We could write instead
```
chmod +x concordium-client*
```

source/mainnet/net/installation/downloads-testnet.rst
```diff
@@ -85,6 +85,14 @@ Concordium Client v4.1.0
 
    - SHA256 checksum of the download: |client-linux-checksum|
 
+   |
+
+   When you run Concordium Client on Linux, run the following command to make the file you download executable:
```
abizjak, line 90:
```suggestion
   Before you can use the downloaded Concordium Client on Linux you have to make the downloaded file executable. Run the following command to make the file you download executable:
```

abizjak (approved):

PR #621, opened 2022-09-30
Update to the invoke host function

### LINKED ISSUE #612: Documentation of the `invoke` host function is incorrect.
The documentation of the `V1` host function `invoke` is not entirely correct. It states
```
Returns

    If the last five bytes are 0 then the call succeeded. In this case, the first bit of the response indicates whether state (of the invoking contract) has changed (1) or not (0) and the remaining 23 bits are the index of the return value that can be used in a call to [get_parameter_size](https://developer.concordium.software/en/mainnet/smart-contracts/references/host-fns.html#concordium.get_parameter_size) and [get_parameter_section](https://developer.concordium.software/en/mainnet/smart-contracts/references/host-fns.html#concordium.get_parameter_section). Otherwise, the call failed and only the forth byte is set. With the value:

    1 if the call failed because of insufficient funds.

    2 if the account to transfer to did not exist.

    3 if the contract to call did not exist.

    4 if the entrypoint did not exist on contract to call.

    5 if it called a V0 contract that failed.

    6 if it called a contract that failed with a runtime error.
```

The part that is not correct. The correct response is documented in `concordium-std`
```
    /// - if the last 5 bytes are 0 then the call succeeded. In this case the
    ///   first bit of the response indicates whether our own state has changed
    ///   (1) or not (0) the remaining 23 bits are the index of the return value
    ///   that can be used in a call to `get_parameter_section` and
    ///   `get_parameter_size`.
    /// - otherwise
    ///   - if the fourth byte is 0 the call failed because of a logic error and
    ///     there is a return value. Bits 1..24 of the response are the index of
    ///     the return value. Bits 32..64 are to be interpreted in two's
    ///     complement and will be a negative number indicating the error code.
    ///   - otherwise only the fourth byte is set.
    ///   - if it is 1 then call failed due to transfer of non-existent amount
    ///   - if it is 2 then the account to transfer to did not exist
    ///   - if it is 3 then the contract to invoke did not exist
    ///   - if it is 4 then the entrypoint did not exist
    ///   - if it is 5 then sending a message to V0 contract failed.
    ///   - if it is 6 then invoking a contract failed with a runtime error
    ///   - no other values are possible
```

dg-concordium:
## Purpose

Add/correct information about the invoke host function for v1 smart contracts

## Changes

Added missing information.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

Closes #612

### DIFF
--- a/source/mainnet/smart-contracts/references/host-fns.rst
+++ b/source/mainnet/smart-contracts/references/host-fns.rst
@@ -351,7 +351,7 @@ Functions only accessible for smart contract receive functions.
             *invoking* contract) has changed (``1``) or not (``0``) and the
             remaining 23 bits are the index of the return value that can be used
             in a call to |get_parameter_size|_ and |get_parameter_section|_.
-            If the fourth byte is 0 the call failed because of a logic error and
+            If the bits 25..32 are all zero the call failed because of a logic error and
             there is a return value. Bits 1..24 of the response are the index of
             the return value. Bits 32..64 are to be interpreted in two's
             complement and will be a negative number indicating the error code.
--- a/source/mainnet/smart-contracts/references/host-fns.rst
+++ b/source/mainnet/smart-contracts/references/host-fns.rst
@@ -351,7 +351,11 @@ Functions only accessible for smart contract receive functions.
             *invoking* contract) has changed (``1``) or not (``0``) and the
             remaining 23 bits are the index of the return value that can be used
             in a call to |get_parameter_size|_ and |get_parameter_section|_.
-            Otherwise, the call failed and only the forth byte is set. With the value:
+            If the fourth byte is 0 the call failed because of a logic error and
+            there is a return value. Bits 1..24 of the response are the index of
+            the return value. Bits 32..64 are to be interpreted in two's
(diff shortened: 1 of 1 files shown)


dg-concordium:
@abizjak I left some of the original wording of the values because they make more sense to the average reader, e.g. insufficient funds instead of non-existent amount. But we can change them if you feel strongly about them.

source/mainnet/smart-contracts/references/host-fns.rst
```diff
@@ -351,7 +351,11 @@ Functions only accessible for smart contract receive functions.
             *invoking* contract) has changed (``1``) or not (``0``) and the
             remaining 23 bits are the index of the return value that can be used
             in a call to |get_parameter_size|_ and |get_parameter_section|_.
-            Otherwise, the call failed and only the forth byte is set. With the value:
+            If the fourth byte is 0 the call failed because of a logic error and
```
abizjak, line 354: Maybe use bit indexing here as well.
```suggestion
            If the bits 25..32 are all zero the call failed because of a logic error and
```

abizjak (approved):

PR #652, opened 2022-10-24
Tokens in wallets

dg-concordium:
## Purpose

Document the addition of CIS2 tokens in wallets

## Changes

Created topic for tokens in wallets.  For now centering on BW but with MW to follow. Added topic to index.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


source/mainnet/net/guides/recover-wallet.rst
```diff
@@ -22,11 +22,23 @@ In case you get a new device or need to restore your existing device, you can re
...
+        :width: 25%
+
     .. Note::
 
         You cannot recover a wallet created in Mainnet in Testnet, and vice versa. Once you have recovered on the correct network, you can :ref:`switch between mainnet and testnet<switch-network>`.
         If you use both networks, you can recover on the other net after setup, through the settings page.
 
+    .. Warning::
+
+        If you had previously had tokens in your wallet, these will not be recovered. You must add the tokens again.
```
soerenbf, line 40: Is this phrasing scary or is it just me? Can sound a bit like you have lost your tokens, if you don't know any better?

Maybe something like `If you had previously configured your wallet to be able to view selected tokens, this will not be recovered. You must add the tokens again`

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+    To access tokens and token transactions, click Tokens |token-bw|. Tabs appear for tokens.
+
+    .. dropdown:: Add tokens
+
+        #. Click **Manage**.
+
+            .. image:: ../images/browser-wallet/manage-tokens.png
+                :width: 50%
+
+        #. Enter a smart contract address on which to look for tokens. Click **Choose contract**. If you get an error that says **Invalid contract address** this means that the smart contract doesn’t exist, the smart contract doesn’t live up to the CIS-2 standard, or the smart contract doesn’t contain any tokens.
```
soerenbf, line 21: "smart contract address" -> "smart contract index"

Also we don't give an error saying "Invalid contract address", but rather different errors depending on the issue:

- 'No contract found on index',
- 'Contract does not support CIS-0',
- 'Contract does not support CIS-2',
- 'No tokens found in contract',

That might be too much to fit in there though.

```suggestion
        #. Enter a smart contract index on which to look for tokens. Click **Choose contract**. If you get an error, this means that the smart contract doesn’t exist, the smart contract doesn’t live up to the CIS-2 standard, or the smart contract doesn’t contain any tokens.
```

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+
+            .. image:: ../images/browser-wallet/add-token-contract.png
+                :width: 50%
+
+        #. Select the token(s) you want to add. You can click on a token to see more information about it and then click **Back to list** to return to the list to select tokens. Click **Update tokens** once you have made your selection(s).
```
soerenbf, line 26: You can't click back to list, but rather click the close icon to close the detail view and go back to viewing the list of tokens.

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+            .. image:: ../images/browser-wallet/select-tokens.png
+                :width: 50%
+
+        .. Note::
+
+            You can add NFTs that are not owned by your account to your wallet. This does not mean that you have ownership of the NFT.
+
+        Once the tokens are added you can see them on the relevant tab.
+
+        An easy way to see how this works is to use the `wCCD example dApp hosted by Concordium <https://wccd.testnet.concordium.com/>`_. Connect it to your |bw| and convert some CCD to wrapped CCD. When you do this, you get the a smart contract number. You can then use this smart contract number to search for tokens on the contract and add them to your |bw|.
```
soerenbf, line 37: Maybe change "smart contract number" to "smart contract index" to stay consistent with the previous section pertaining to adding tokens.

source/mainnet/net/guides/recover-wallet.rst
shjortConcordium: `... these will not be recovered automatically recovered. You must add the tokens to view them again.`
Maybe?

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+
+    .. dropdown:: Inspect tokens
+
+        Click the tab for the token you want to inspect. Then click on the token to see details for that token.
```
shjortConcordium, line 41:
```suggestion
        Click the tab for the type of token that you want to inspect. Then click on the token to see details for that token.
```

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+
+        You can click **Show raw metadata** to see the raw metadata for the token. You can also click **Don't show token in wallet** if you don't want the token to appear in your wallet.
+
+    .. dropdown:: Send tokens
+
+        #. Click Send |send-bw| in the Accounts overview.
+
+        #. Click the field for what to send.
+
+            .. image:: ../images/browser-wallet/send-tokens-select.png
```
shjortConcordium, line 51: This view was changed, actually. (to make it more obvious that the highlighted box can be clicked)

shjortConcordium: Screenshot from 2022-11-16 13-49-18

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+                :width: 25%
+
+            .. image:: ../images/browser-wallet/send-tokens-select2.png
+                :width: 25%
+
+        #. In the amount field enter the amount of the token you want to send. The number of decimal places depends on the token.
+
+        #. Add the receiver address. Paste the address, use the address book, or scan a QR code to add the address.
```
shjortConcordium, line 59: There are no address book and it can't scan QR codes. :cry:

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+
+            .. note::
+                If the token is a CCD token, you have the option to add a memo to the transaction. Remember that memos increase the transaction fee.
```
shjortConcordium, line 62: We can't do this either. (in the browser wallet :thinking: )

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,80 @@
...
+
+        #. Click **Continue**.
+
+        #. Review the transaction. When satisfied, click **Send**.
+
+        #. Click **Finish**.
+
+.. Warning::
+
+    If you had previously had tokens in your wallet, these will not be recovered. You must add the tokens again.
```
shjortConcordium, line 72: Should this be here, it sounds weird in this context. :thinking:

dg-concordium: Because each of the actions is in a dropdown, it would actually fall just below "You can add, inspect, and send tokens in the |bw|" at the top. It opens collapsed by default.

dg-concordium: image

dg-concordium: It is also because the same message will presumably apply to all wallets, not just BW. I can move it above the dropdown list.

source/mainnet/net/guides/tokens.rst
dg-concordium: It's still so subtle that most users won't even notice it. So I will likely keep the red box around the field.

source/mainnet/net/guides/tokens.rst
dg-concordium: Oh, I think I see what you mean. Why is it in tokens and also in recover wallet. I think it's better to state these things multiple times, lots of places.

soerenbf (approved):
Looks good to me 👍

source/mainnet/net/guides/tokens.rst
```diff
@@ -0,0 +1,82 @@
...
+
+            .. image:: ../images/browser-wallet/add-token-contract.png
+                :width: 50%
+
+        #. Select the token(s) you want to add. You can click on a token to see more information about it and then click **X** to return to the list to select tokens. You can also search by token ID at the top of the screen. Click **Update tokens** once you have made your selection(s).
+
+            .. image:: ../images/browser-wallet/select-tokens.png
+                :width: 50%
+
+        If you have purchased tokens on a marketplace, you can add the tokens to your wallet from the dApp or service so you do not have to go through the steps above. The screen below appears when the purchase is triggered and your wallet is connected. Click **Add selected** to add the selected tokens to your wallet.
```
shjortConcordium, line 31: The issue with this is that it makes it sound like any marketplace will do this, and not a "optional" thing that they may implement, and decide themselves when to trigger. :thinking: 
```suggestion
        If you have purchased tokens on a marketplace, it can suggest tokens to be added directly without going through the steps above. In this case the screen below appears when your wallet is connected and the service suggests the tokens to be added. Click **Add selected** to add the selected tokens to your wallet.
```
Something like this?

PR #670, opened 2022-11-15
Add limits for parameter, return_value, and logs

DOBEN:
## Purpose

Add the limits to documentation regarding the parameter size limit, return_value size limit, and number of log items.

Note:  This should not be merged before the protocol update is in effect on testnet.

### DIFF
--- a/source/mainnet/smart-contracts/general/contract-instances.rst
+++ b/source/mainnet/smart-contracts/general/contract-instances.rst
@@ -158,6 +158,8 @@ Logging an event has an associated cost, similar to the cost of writing to the
 contract's state. In most cases it would only make sense to log a few bytes to
 reduce cost.
 
+There is no limit to the number of logs per invocation (apart from energy).
+
 .. _contract-instance-operations:
 
 Limits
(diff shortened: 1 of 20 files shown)


source/mainnet/smart-contracts/general/contract-instances.rst
```diff
@@ -152,6 +160,15 @@ reduce cost.
...
 .. _contract-instance-operations:
 
+Limits
+======
+
+When initializing or invoking a smart contract, the following limits are enforced:
+
+- The maximum parameter size is 65535 Bytes.
+- Up to 4 billion log items per execution.
+- The maximum return value size is 4 GigaBytes.
```
Bargsteen, line 170: I am not sure that we should mention these limits as you will hit the energy limit well before that. At least we should mention the energy limit.

source/mainnet/smart-contracts/general/contract-instances.rst
```diff
@@ -152,6 +160,15 @@ reduce cost.
 
 .. _contract-instance-operations:
 
+Limits
+======
+
+When initializing or invoking a smart contract, the following limits are enforced:
```
Bargsteen, line 166:
```suggestion
When initializing, updating, or invoking a smart contract, the following limits are enforced:
```

Bargsteen (changes requested):

source/mainnet/smart-contracts/general/contract-instances.rst
DOBEN: What is the current energy limit for the init/receive function?

dg-concordium (approved):

source/mainnet/smart-contracts/general/contract-instances.rst
```diff
@@ -152,6 +160,15 @@ reduce cost.
...
+
+- The maximum parameter size is 65535 Bytes.
+- Up to 4 billion log items per execution.
```
abizjak, line 169: We can be more precise here, it is `u32::MAX` logs.

source/mainnet/smart-contracts/general/contract-instances.rst
abizjak, line 170: To be more precise it is u32::MAX bytes. 

I think we should have the limits, but still state that they cannot be reached in practice due to NRG bounds.

abizjak (approved):

source/mainnet/smart-contracts/general/contract-instances.rst
```diff
@@ -152,6 +160,20 @@ reduce cost.
...
 
+Limits
+======
+
+When initializing, updating, or invoking a smart contract, the following limits are enforced:
+
+- The maximum input parameter size is 65535 Bytes (u16::MAX).
+- The energy limit in a block is 3 million NRG (energy). If only one transaction is in the block, the transaction can consume up to 3 million NRG (energy).
+- The maximim number of log items is u32::MAX per execution.
+- The maximum return value size is u32::MAX Bytes.
```
Bargsteen, line 171:
```suggestion
- The maximum input parameter size is 65535 Bytes (``u16::MAX``).
- The energy limit in a block is 3 million NRG (energy). If only one transaction is in the block, the transaction can consume up to 3 million NRG (energy).
- The maximum number of log items is ``u32::MAX`` per execution.
- The maximum return value size is ``u32::MAX`` Bytes.
```

Bargsteen (approved):
LGTM. But should not be merged before the protocol update is in effect on testnet.

PR #722, opened 2023-01-10
The DAO reentrancy example

### LINKED ISSUE #435: Explain reentrancy issues with an example
**Task description**
There is currently a warning in the smart contract unit test guide about problems with reentrancy, but it is not very elaborate.
Explain the reentrancy issue with an example. 
The example could be similar to this counter-notify contract:

``` rust
#![cfg_attr(not(feature = "std"), no_std)]
use concordium_std::*;

type State = u64;

#[init(contract = "counter-notify")]
#[inline(always)]
fn contract_init<S: HasStateApi>(
    _ctx: &impl HasInitContext,
    _state_builder: &mut StateBuilder<S>,
) -> InitResult<State> {
    Ok(0u64)
}

#[receive(contract = "counter-notify", name = "just-increment", mutable)]
fn just_increment<S: HasStateApi>(
    _ctx: &impl HasReceiveContext,
    host: &mut impl HasHost<State, StateApiType = S>,
) -> ReceiveResult<()> {
    *host.state_mut() += 1;
    Ok(())
}

#[receive(contract = "counter-notify", name = "increment-and-notify", mutable)]
fn increment_and_notify<S: HasStateApi>(
    ctx: &impl HasReceiveContext,
    host: &mut impl HasHost<State, StateApiType = S>,
) -> ReceiveResult<bool> {
    let (contract, entrypoint): (ContractAddress, OwnedEntrypointName) =
        ctx.parameter_cursor().get()?;

    // Increment counter
    *host.state_mut() += 1;
    let preinvoke_count = *host.state();

    // Notify a contract about the new counter value.
    host.invoke_contract(
        &contract,
        Parameter(&to_bytes(&preinvoke_count)),
        entrypoint.as_entrypoint_name(),
        Amount::zero(),
    )
    .unwrap_abort();

    // This will be false if the reentrancy has occurred.
    Ok(preinvoke_count == *host.state())
}

#[concordium_cfg_test]
mod tests {
    use super::*;
    use concordium_std::claim_eq;
    use test_infrastructure::*;

    #[concordium_test]
    fn state_can_change_due_to_reentrancy() {
        let mut ctx = TestReceiveContext::empty();
        let self_address = ContractAddress {
            index:    0,
            subindex: 0,
        };
        let entrypoint_just_increment = OwnedEntrypointName::new_unchecked("just-increment".into());
        let parameter_bytes = to_bytes(&(self_address, entrypoint_just_increment.clone()));
        ctx.set_parameter(&parameter_bytes);
        ctx.set_self_address(self_address);

        let mut host = TestHost::new(0u64);

        // We are simulating reentrancy with this mock because we mutate the state.
        host.setup_mock_entrypoint(
            self_address,
            entrypoint_just_increment,
            MockFn::new_v1(|_parameter, _amount, _balance, state: &mut State| {
                *state += 1;
                Ok((true, ()))
            }),
        );

        let res = increment_and_notify(&ctx, &mut host).expect_report("Calling receive failed.");

        // The count
        claim_eq!(res, false);
    }
}
```

**Sub-tasks**
- [ ] Find a suitable example OR revise the counter-notify example
- [ ] Explain the reentrancy problem in the documentation using snippets from the example

annenkov:
## Purpose

Add a reentrancy example to the unit testing section

Closes  #435.

### DIFF
--- a/source/mainnet/smart-contracts/guides/unit-test-contract.rst
+++ b/source/mainnet/smart-contracts/guides/unit-test-contract.rst
@@ -428,7 +428,7 @@ In particular, the example below mimics the original ``withdraw_reentrancy`` cod
         host.set_self_balance(Amount::from_micro_ccd(2));
 
         // Set up a mock entrypoint that calls back to our contract.
-        // The mock emulates the `withdraw_reentrancy` logic to model 
+        // The mock emulates the `withdraw_reentrancy` logic to model
         // a reentrancy attack that will withdraw the sender's balance twice.
         host.setup_mock_entrypoint(
             CONTRACT_ADDRESS,
--- a/source/mainnet/smart-contracts/guides/unit-test-contract.rst
+++ b/source/mainnet/smart-contracts/guides/unit-test-contract.rst
@@ -353,10 +353,10 @@ You should watch out for *reentrancy problems*, which can occur when calls to
     // *host.state() and state_copy might not be equal any more due to reentrancy.
     do_something_with(state_copy);
 
-Another example of reentracny is when the state is *not* updated properly before making an external call.
-This can lead to reentrant calls that pass some validation that is based on the current state, while these calls should fail.
+Another example of reentrancy is when the state is *not* updated properly before making an external call.
+This can lead to reentrant calls that pass some validation that is based on the current state, even though these calls should fail.
 The classic example of such a security issue is the Ethereum DAO smart contract that was drained of funds due to the reentrancy vulnerability.
-We show a code snippet that implements a small part similar to the DAO contract that stores balances for arbitrary addresses in a map ``StateMap<Address, Amount, S>``.
+Below is a code snippet that implements a small part similar to the DAO contract that stores balances for arbitrary addresses in a map ``StateMap<Address, Amount, S>``.
 The users can request their funds back; if a user is a smart contract, the funds are sent to a specified entrypoint.
 
 .. code-block:: rust
@@ -412,7 +412,7 @@ The problem in the code above is that resetting the sender's balance to zero hap
 The sender's balance in the *contract state* is used to determine how much funds should be transferred to the sender.
 Since it is not updated, the external contract can make a call back to ``reentrant_withdraw`` and pass the balance validation.
 Testing this behavior with mocks require some insights.
-In particular, we are going to mimic the original ``reentrant_withdraw`` code in the mock entrypoint.
+In particular, the example below mimics the original ``reentrant_withdraw`` code in the mock entrypoint.
 
 .. code-block:: rust
 
@@ -474,9 +474,9 @@ The test fails with the following message:
 
     Incorrect balance: expected Amount { micro_ccd: 1 }, found: Amount { micro_ccd: 0 }
 
-That means that the contract we called has stolen our funds through a reentrant call.
+That means that the contract called has stolen funds through a reentrant call.
 A simple fix to this behavior is to place the highlighted line in ``reentrant_withdraw`` *before* making a call to an external contract.
-In this case we expect the ``reentrant_withdraw`` call to fail because the non-zero balance condition is not longer satisfied in the mock entrypoint.
+In this case the ``reentrant_withdraw`` call will fail because the non-zero balance condition is no longer satisfied in the mock entrypoint.
 
 Testing with state rollbacks
 ============================
--- a/source/mainnet/smart-contracts/guides/unit-test-contract.rst
+++ b/source/mainnet/smart-contracts/guides/unit-test-contract.rst
@@ -342,8 +342,11 @@ state and balance fields:
 Reentrancy
 ----------
 
-You should watch out for *reentrancy problems*, which can occur when calls to
-``invoke_contract`` end up updating the state of your own contract.
+When invoking another smart contract, you give away control to that contract in the middle of execution.
+The external contract can, for example, call back entrypoints of your contract.
+This behavior is called *reentrancy* and is well-known from concurrency: a procedure can be interrupted in the middle of its execution, called again, and then resume execution.
+
+The state of your contract might not be the same before and after ``invoke_contract``, since the contract you call can update the state of your own contract.
 
 .. code-block:: rust
 
@@ -353,7 +356,7 @@ You should watch out for *reentrancy problems*, which can occur when calls to
     // *host.state() and state_copy might not be equal any more due to reentrancy.
     do_something_with(state_copy);
 
-Another example of reentrancy is when the state is *not* updated properly before making an external call.
+Consider a concrete example of reentrancy when the state is *not* updated properly before making an external call.
 This can lead to reentrant calls that pass some validation that is based on the current state, even though these calls should fail.
 The classic example of such a security issue is the Ethereum DAO smart contract that was drained of funds due to the reentrancy vulnerability.
 Below is a code snippet that implements a small part similar to the DAO contract that stores balances for arbitrary addresses in a map ``StateMap<Address, Amount, S>``.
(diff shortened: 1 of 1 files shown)


source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
+    host.invoke_contract(...);
+
+    // *host.state() and state_copy might not be equal any more due to reentrancy.
+    do_something_with(state_copy);
+
+Another example of reentracny is when the state is *not* updated properly before making an external call.
+This can lead to reentrant calls that pass some validation that is based on the current state, while these calls should fail.
+The classic example of such a security issue is the Ethereum DAO smart contract that was drained of funds due to the reentrancy vulnerability.
+We show a code snippet that implements a small part similar to the DAO contract that stores balances for arbitrary addresses in a map ``StateMap<Address, Amount, S>``.
+The users can request their funds back; if a user is a smart contract, the funds are sent to a specified entrypoint.
```
abizjak, line 360: I think this description, and below, is too specific and does not state the key point clearly. 

Re-entrancy is a special case. The point we need to drive across is that invoking another contract passes control out of the current execution context. 

We should also point out that this is not a "novel smart contract issue". It is a standard issue you get when you have, e.g., concurrency. 

So instead of pointing out that they should watch for re-entrancy problems, we should tell them the "positive" thing that they should always make sure to record state updates before doing things that are outside of the contract's control. This way any entrypoint that is ever called will have "clean" state to work with.

We can then list (a simplified) version of the DAO example as a non-example.

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
+
+.. code-block:: rust
+
+    let state_copy = *host.state();
+    host.invoke_contract(...);
+
+    // *host.state() and state_copy might not be equal any more due to reentrancy.
+    do_something_with(state_copy);
+
+Another example of reentracny is when the state is *not* updated properly before making an external call.
```
dg-concordium, line 356:
```suggestion
Another example of reentrancy is when the state is *not* updated properly before making an external call.
```

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
+This can lead to reentrant calls that pass some validation that is based on the current state, while these calls should fail.
```
dg-concordium, line 357:
```suggestion
This can lead to reentrant calls that pass some validation that is based on the current state, even though these calls should fail.
```

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
+The classic example of such a security issue is the Ethereum DAO smart contract that was drained of funds due to the reentrancy vulnerability.
+We show a code snippet that implements a small part similar to the DAO contract that stores balances for arbitrary addresses in a map ``StateMap<Address, Amount, S>``.
```
dg-concordium, line 359:
```suggestion
Below is a code snippet that implements a small part similar to the DAO contract that stores balances for arbitrary addresses in a map ``StateMap<Address, Amount, S>``.
```

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
+        }
+
+        Ok(())
+    }
+
+The problem in the code above is that resetting the sender's balance to zero happens *after* the call to an external contract is completed.
+The sender's balance in the *contract state* is used to determine how much funds should be transferred to the sender.
+Since it is not updated, the external contract can make a call back to ``reentrant_withdraw`` and pass the balance validation.
+Testing this behavior with mocks require some insights.
+In particular, we are going to mimic the original ``reentrant_withdraw`` code in the mock entrypoint.
```
dg-concordium, line 415:
```suggestion
In particular, the example below mimics the original ``reentrant_withdraw`` code in the mock entrypoint.
```

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
-   .. code-block:: rust
+.. code-block:: text
 
-      let state_copy = *host.state();
-      host.invoke_contract(...);
+    Incorrect balance: expected Amount { micro_ccd: 1 }, found: Amount { micro_ccd: 0 }
 
-      // *host.state() and state_copy might not be equal any more due to reentrancy.
-      do_something_with(state_copy);
+That means that the contract we called has stolen our funds through a reentrant call.
```
dg-concordium, line 477:
```suggestion
That means that the contract called has stolen funds through a reentrant call.
```

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
+A simple fix to this behavior is to place the highlighted line in ``reentrant_withdraw`` *before* making a call to an external contract.
+In this case we expect the ``reentrant_withdraw`` call to fail because the non-zero balance condition is not longer satisfied in the mock entrypoint.
```
dg-concordium, line 479:
```suggestion
In this case the ``reentrant_withdraw`` call will fail because the non-zero balance condition is no longer satisfied in the mock entrypoint.
```

dg-concordium (approved):
A few minor changes. I generally avoid using first person in the documentation (we, our, I, my, etc.).

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -339,18 +339,144 @@ state and balance fields:
...
+        // We emulate the `reentrant_withdraw` logic here.
+        host.setup_mock_entrypoint(
+            CONTRACT_ADDRESS,
+            OwnedEntrypointName::new_unchecked("reentrant_withdraw".to_string()),
+            MockFn::new_v1(|_parameter, _amount, balance, state: &mut State<_>| {
+                // We cannot call `invoke_contract` inside this mock, but we have
+                // access to the `balance`, which is the balance of the contract
+                // making this invocation. So we can simulate invoking withdraw by
+                // subtracting the sender's amount stored in the contract state
+                // from the balance.
```
Bargsteen, line 437: Perhaps we could mention that we are simulation a reentrancy attack, where we will withdraw the balance twice.

source/mainnet/smart-contracts/guides/unit-test-contract.rst
Bargsteen: I think it makes sense to mention that it is a general concurrency issue, as long as we don't write it in a way that requires much knowledge about concurrency by the readers.

Bargsteen (approved):
Nice work 😊

source/mainnet/smart-contracts/guides/unit-test-contract.rst
annenkov: I'll try to generalise the problem statement and refer to concurrency. I'll make a more detailed description of reentrancy in best practices and cross-reference it with testing.

I've created an issue for adding the full code of the DAO example https://github.com/Concordium/concordium-rust-smart-contracts/issues/215

source/mainnet/smart-contracts/guides/unit-test-contract.rst
annenkov: I added a line about this before the mock entrypoint definition.

source/mainnet/smart-contracts/guides/unit-test-contract.rst
```diff
@@ -342,8 +342,11 @@ state and balance fields:
 Reentrancy
 ----------
 
-You should watch out for *reentrancy problems*, which can occur when calls to
-``invoke_contract`` end up updating the state of your own contract.
+When invoking another smart contract, you give away control to that contract in the middle of execution.
+The external contract can, for example, call back entrypoints of your contract.
+This behavior is called *reentrancy* and is well-known from concurrency: a procedure can be interrupted in the middle of its execution, called again, and then resume execution.
+
+The state of your contract might not be the same before and after ``invoke_contract``, since the contract you call can update the state of your own contract.
```
abizjak, line 349: Can we be a bit more precise. In other places we are adamant about the fact that only you and you alone can update the contract state. But here it now says "the contract you call can update the state ..." which is technically incorrect.

Can you please rephrase this so that it is both technically correct, and fits into the narrative we try to have.

PR #737, opened 2023-01-24
Testnet service node

dg-concordium:
## Purpose

Inform users about availability of a testnet service node.

## Changes

Added information in a couple of discreet locations.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.

### DIFF
--- a/source/mainnet/net/guides/developer-page.rst
+++ b/source/mainnet/net/guides/developer-page.rst
@@ -28,7 +28,7 @@ Official resources
 
 - See :ref:`dApp examples<dapp-examples>` for a list of dApp examples and their resources.
 
-- For testing purposes, an Ubuntu node is available on testnet to use when testing smart contracts and dApps. The address is node.testnet.concordium.com on port 10000 (GRPCv1) and port 20000 (gRPCv2 and gRPC-web). Node-related methods such as NodeInfo and Shutdown have been disabled. This node is periodically maintained by Concordium. The status of this node is available on the `Testnet status page <https://status.testnet.concordium.software>`__.
+- For testing purposes, an Ubuntu node is available on testnet to use when testing smart contracts and dApps. You can use this node for API calls of chain methods only with GRPC v1, gRPC v2, and gRPC web. The address is node.testnet.concordium.com on port 10000 (GRPCv1) and port 20000 (gRPCv2 and gRPC-web). This node is periodically maintained by Concordium. The status of this node is available on the `Testnet status page <https://status.testnet.concordium.software>`__.
 
 - Block explorers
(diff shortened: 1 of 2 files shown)


source/mainnet/net/guides/developer-page.rst
```diff
@@ -28,6 +28,8 @@ Official resources
 
 - See :ref:`dApp examples<dapp-examples>` for a list of dApp examples and their resources.
 
+- For testing purposes, an Ubuntu node is available on testnet to use when testing smart contracts and dApps. You can use this node for API calls of chain methods only with GRPC v1, gRPC v2, and gRPC web. The address is node.testnet.concordium.com on port 10000 (GRPCv1) and port 20000 (gRPCv2 and gRPC-web). This node is periodically maintained by Concordium. The status of this node is available on the `Testnet status page <https://status.testnet.concordium.software>`__.
```
abizjak, line 31: Why does it matter that it is an Ubuntu node? I recommend removing this. What is important is that it has grpc... exposed on ports 10000 and 20000 as is stated.

source/mainnet/smart-contracts/tutorials/setup-env.rst
```diff
@@ -230,6 +230,10 @@ Remember you are working on the testnet. Check if your node collector is up and
 .. image:: images/node-collector.png
     :width: 100%
 
+.. Note::
+
+    For testing purposes, a node is available on testnet to use when testing smart contracts and dApps. You can use this node for API calls of chain methods only with GRPC v1, gRPC v2, and gRPC web. The address is node.testnet.concordium.com on port 10000 (GRPCv1) and port 20000 (gRPCv2 and gRPC-web). This node is periodically maintained by Concordium. The status of this node is available on the `Testnet status page <https://status.testnet.concordium.software>`__.
```
abizjak, line 235: I don't understand the phrasing "periodically maintained". Could this be rephrased?

WHat is the intended meaning? That we don't guarantee any specific uptime @fdibat ?

dg-concordium: I can rephrase, but the intention is to let the users know that there is no guarantee on uptime or that they can expect immediate assistance if the node isn't working.

dg-concordium: Something like this? "This node is maintained by Concordium, but Concordium does not guarantee availability or uptime."

fdibat: @abizjak we should not commit to a specific uptime. We will fix if something happen but it will also depend from other priorities and from the amount of "troubles" we will get from it. I would remove the word "uptime" completely from it. "This node is maintained by Concordium, but Concordium does not guarantee any specific availability".

abizjak (approved):

PR #763, opened 2023-02-14
Smart contract costs best practices

dg-concordium:
## Purpose

Describe various cost reduction measures that could be employed when writing and deploying smart contracts.

## Changes

- Reintroduce the index page for best practices
- Add a page for cost reductions
- The sphinx containers were updated yesterday, so `latest` now uses python `3.12`, which means that the CI jobs here didn't work. I've updated the python version in the pip file and also the lock file.

## Checklist

      hard-to-understand areas.


dg-concordium:
@Bargsteen Is it also possible to do something with this older pull request in the "Iterate on smart contract documentation" work?

Bargsteen:
@DOBEN @limemloh The preview is now working, so please take a look when you have the time ;)

source/mainnet/smart-contracts/best-practices/costs.rst
```diff
@@ -0,0 +1,75 @@
...
+For very complex contracts it may be beneficial to run benchmarks to see
+whether ``bump_alloc`` is the best option. See the Rust `allocator <https://doc.rust-lang.org/std/alloc/index.html#the-global_allocator-attribute>`_
+documentation for more context and details on using custom allocators.
+
+There are other allocators available, for example ``dlmalloc``, that can be used instead.
+
+Use State-* types for maps, sets, and boxes
+===========================================
+
+If your contract uses maps, sets, or boxes, then *strongly* consider using the alternatives prefixed with ``State`` from ``concordium-std``. For example |StateMap|_, |StateBTreeMap|_, |StateSet|_, |StateBTreeSet|_, or |StateBox|_.
```
limemloh, line 53: Maybe a minor detail, but they are probably only more cost-efficient if you need to persist this collection across calls and don't need to read the entire collection on every call, otherwise I would expect rust-std types to be more cost-efficient, at least in some scenarios :shrug: 
But I believe this is true for a majority of the use cases, so I am okay with adding this

limemloh (approved):

source/mainnet/smart-contracts/best-practices/costs.rst
```diff
@@ -0,0 +1,75 @@
...
+The ``State`` versions efficiently utilize the underlying smart contract state for a vastly improved performance.
+The underlying the smart contract state is a tree of byte arrays.
```
DOBEN, line 55:
```suggestion
The underlying smart contract state is a tree of byte arrays.
```

DOBEN (approved):

source/mainnet/smart-contracts/best-practices/costs.rst
Bargsteen: Good point, I've added "and stores"

PR #825, opened 2023-05-23
Run a local chain

vikt0r0:
## Purpose

Add a minimal example showing how to run a local chain.

## Changes

- Add guide `source/mainnet/net/guides/run-local-chain.rst`.
- Update `source/mainnet/net/index.rst`.

## Checklist

      hard-to-understand areas.


source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    stake = "500000000000000"
+    template = "baker"
+    identityProvider = 0
+    numKeys = 1
+    threshold = 1
+    repeat = 1 # Changed from 5
+
+.. Note::
+
+    Note that the staked amount needed to participate in finalization committee is some fraction of the total amount of existing CCD, defined by the value of the ``capitalBound`` key in the configurations file. The total amount is the sum of the balances of all the baker and foundation accounts specified in the genesis configuration file.
```
vikt0r0, line 98: I am not really sure about the `capitalBound` claim, but I feel this may be important. Or I should just state what it is in this case.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+Run a local chain
+=================
+
+This guide details how to run your own instance of the Concordium blockchain. This is useful when developing and testing smart contracts. Running your own chain also lets you control various aspects such as the genesis parameters, anonymity revokers, identity providers and foundation accounts.
+
+The chain is run by a network of baker nodes that bake and finalize blocks. In the following minimal example you will set up a network comprised of a single baker node that runs *locally* on your system. Note, however, that the concepts demonstrated here equally apply to any number of baker nodes configured in a LAN or WAN setting.
+
+.. Note::
+
+   The commands shown in the following are interpreted in a ``bash``-like environment.
```
vikt0r0, line 13: Probably unncessary...

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    stake = "500000000000000"
+    template = "baker"
+    identityProvider = 0
+    numKeys = 1
+    threshold = 1
+    repeat = 1 # Changed from 5
+
+.. Note::
+
+    Note that the staked amount needed to participate in finalization committee is some fraction of the total amount of existing CCD, defined by the value of the ``capitalBound`` key in the configurations file. The total amount is the sum of the balances of all the baker and foundation accounts specified in the genesis configuration file. In this particular example the stake is sufficient for baking.
```
vikt0r0, line 98: Not sure about this `capitalBound` claim.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+
+The ``--no-bootstrap`` option lets the node know not to connect to a bootstrapper node for retrieving peers since no peers are in this network so this is not relevant. The ``--grpc2-listen-port`` specifies the port to listen on for Node GRPC V2 API handshakes. We will use this interface to communicate with the node. The ``--data-dir`` and ``--config-dir`` specifies the working directories of the node instance, where its state and configuration is stored. The ``--baker-credentials-file`` instructs the node to run as the baker specified in the supplied credentials file, in this case your generated baker credentials output from the ``genesis-creator`` tool.
+
+.. Note::
+
+    If more baker credentials are generated, more bakers can be spun up by replacing the arguments specified by the ``--baker-credentials-file``, ``--config-dir`` and ``--data-dir`` options accordingly. If there is no bootstrapper node, it will have to be instructed to manually connect to one another by specifying the IP address and port of the other node(s) using the ``--connect-to`` option. Note that nodes running on the same network interfaces must specify a disjoint set of ports.
```
vikt0r0, line 148: Maybe just get rid of this note?

source/mainnet/net/index.rst
```diff
@@ -53,6 +53,7 @@ index
    guides/developer-page
    nodes/node-requirements
    references/concordium-client
+   guides/run-local-chain
```
vikt0r0, line 56: When I run this locally, this hyperlink only shows up in the sidebar when on the `guides/developer-page` (i.e. it is not visible when I visit `node-requirements` or `references/concordium-client`). I am not sure what causes this, or if it is just an issue local to my system.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+
+   The commands shown in the following are interpreted in a ``bash``-like environment.
+
+Prerequisites
+=============
+You need a working installation of the Concordium Node distribution version 5 or above, instances of which will be used to run the baker node(s), :ref:`genesis block<glossary-genesis-block>` data which specifies your local chain and sets of credentials for baker accounts of the chain. The number of sets of credentials needed will thus depend on the desired number of bakers in the network, which in this example is 1.
+
+Installing the node distribution
+--------------------------------
+Concordium Node releases currently exist for Ubuntu, MacOS, Windows and Docker. See the :ref:`Node Requirements<node-requirements>` section for information on system requirements and detailed instructions on how to obtain, run and manage a node. You need the ``concordium-node`` binary supplied with the distributions in your path to run a baker. Upon successful installation of the distribution verify that the the binary is  in your path:
```
dg-concordium, line 21:
```suggestion
Concordium Node releases exist for Ubuntu, MacOS, Windows, and Docker. See the :ref:`Node Requirements<node-requirements>` section for information on system requirements and detailed instructions on how to obtain, run, and manage a node. To run a baker, you need the ``concordium-node`` binary supplied with the distributions in your path. Upon successful installation of the distribution verify that the binary is in your path:
```

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    concordium_node 5.3.2
+
+.. Note::
+
+   The node distributions provide commands for running and managing nodes on :ref:`Mainnet<glossary-mainnet>` and :ref:`Testnet<glossary-testnet>`. Ignore these since you are targeting your own chain and only need the binary.
+
+
+Generating genesis data and account credentials
+-----------------------------------------------
+You will use the `genesis-creator <https://github.com/Concordium/concordium-misc-tools/tree/main/genesis-creator>`_ tool for generating genesis block data and credentials for the foundation and (initial) baker accounts.
```
dg-concordium, line 35:
```suggestion
Use the `genesis-creator <https://github.com/Concordium/concordium-misc-tools/tree/main/genesis-creator>`_ tool to generate genesis block data and credentials for the foundation and (initial) baker accounts.
```

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+.. code-block:: console
+
+    $ protoc --version
+    libprotoc 3.21.12
+    $ rustc --version
+    rustc 1.68.2 (9eb3afe9e 2023-03-27)
+
+.. Note::
+
+    The recommended method to install Rust is using ``rustup``. After installing ``rustup`` Rust toolchain is then easily be installed by issuing ``rustup toolchain install 1.68``.
```
dg-concordium, line 51:
```suggestion
    The recommended method to install Rust is using ``rustup``. After installing ``rustup``, the Rust toolchain is then easily installed by issuing ``rustup toolchain install 1.68``.
```

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+
+* the initial protocol version
+* cryptographic parameters
+* anonymity revokers
+* identity providers
+* foundation accounts
+* keys for updating the chain
+* various parameters for the genesis
+
+It furthermore specifies where to save the output which is used to invoke the node binary. Many of these options are not relevant when testing smart constracts and the easiest way to get started is to piggyback on one of the examples are found in the ``./examples`` folder. In the following, you will use the file ``./examples/genesis5.toml`` and modify it slightly. Inspecting the file reveals that it specifies an initial protocol version of 5, to output credentials for 5 baker accounts, 2 foundation accounts and the genesis time set to the system time at generation. It also specifies 5 seconds as the average time per block. Further inpection of the table at the ``accounts`` key of the file reveals that the bakers each have an initial balance of 10^15 microCCD and a stake of 5 * 10^14 microCCD. Change this section s.t. just one baker credential is produced by setting the value of the ``repeat`` key to 1 as follows:
```
dg-concordium, line 82:
```suggestion
Furthermore, it specifies where to save the output that is used to invoke the node binary. Many of these options are not relevant when testing smart contracts and the easiest way to get started is to piggyback on one of the examples in the ``./examples`` folder. In the following, you will use the file ``./examples/genesis5.toml`` and modify it slightly. Inspecting the file reveals that it specifies an initial protocol version of 5, to output credentials for 5 baker accounts, 2 foundation accounts, and the genesis time set to the system time at generation. It also specifies 5 seconds as the average time per block. Further inspection of the table at the ``accounts`` key of the file reveals that the bakers each have an initial balance of 10^15 microCCD and a stake of 5 * 10^14 microCCD. Change this section s.t. just one baker credential is produced by setting the value of the ``repeat`` key to 1 as follows:
```

dg-concordium: Also, in the last sentence, what is "s.t." meant to stand for?

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    stake = "500000000000000"
+    template = "baker"
+    identityProvider = 0
+    numKeys = 1
+    threshold = 1
+    repeat = 1 # Changed from 5
+
+.. Note::
+
+    Note that the staked amount needed to participate in finalization committee is some fraction of the total amount of existing CCD, defined by the value of the ``capitalBound`` key in the configurations file. The total amount is the sum of the balances of all the baker and foundation accounts specified in the genesis configuration file. In this particular example the stake is sufficient for baking.
```
dg-concordium, line 98:
```suggestion
    Note that the staked amount needed to participate in the finalization committee is some fraction of the total amount of existing CCD defined by the value of the ``capitalBound`` key in the configurations file. The total amount is the sum of the balances of all the baker and foundation accounts specified in the genesis configuration file. In this particular example, the stake is sufficient for baking.
```

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    Baker keys will be generated in ./bakers
+    Cryptographic parameter will be generated in ./global
+    The genesis data will be stored in ./genesis.dat
+    The genesis hash will be written to ./genesis_hash
+    There are 2 accounts in genesis, 1 of which are bakers.
+    Genesis time is set to 2023-05-22 15:08:19.803 UTC.
+    Average block time is set to 5000ms.
+    DONE
+
+In particular, the files of interest to you are ``./genesis.dat`` containing the genesis block data and ``./bakers/baker-0-credentials.json`` containing the credentials of the single baker account that was created.
```
dg-concordium, line 119:
```suggestion
In particular, the files of interest to you are ``./genesis.dat``, containing the genesis block data, and ``./bakers/baker-0-credentials.json``, containing the credentials of the single baker account that was created.
```

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+
+    concordium-node \
+      --no-bootstrap= \
+      --grpc2-listen-addr 127.0.0.1 \
+      --grpc2-listen-port 20001 \
+      --data-dir localchain-node-0 \
+      --config-dir localchain-node-0 \
+      --baker-credentials-file bakers/baker-0-credentials.json
+
+The ``--no-bootstrap`` option lets the node know not to connect to a bootstrapper node for retrieving peers since no peers are in this network so this is not relevant. The ``--grpc2-listen-port`` specifies the port to listen on for Node GRPC V2 API handshakes. We will use this interface to communicate with the node. The ``--data-dir`` and ``--config-dir`` specifies the working directories of the node instance, where its state and configuration is stored. The ``--baker-credentials-file`` instructs the node to run as the baker specified in the supplied credentials file, in this case your generated baker credentials output from the ``genesis-creator`` tool.
```
dg-concordium, line 144:
```suggestion
The ``--no-bootstrap`` option lets the node know not to connect to a bootstrapper node for retrieving peers since no peers are in this network so this is not relevant. The ``--grpc2-listen-port`` specifies the port to listen on for Node GRPC V2 API handshakes. You will use this interface to communicate with the node. The ``--data-dir`` and ``--config-dir`` specify the working directories of the node instance, where its state and configuration are stored. The ``--baker-credentials-file`` instructs the node to run as the baker specified in the supplied credentials file, in this case, your generated baker credentials output from the ``genesis-creator`` tool.
```

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+
+.. Note::
+
+    If more baker credentials are generated, more bakers can be spun up by replacing the arguments specified by the ``--baker-credentials-file``, ``--config-dir`` and ``--data-dir`` options accordingly. If there is no bootstrapper node, it will have to be instructed to manually connect to one another by specifying the IP address and port of the other node(s) using the ``--connect-to`` option. Note that nodes running on the same network interfaces must specify a disjoint set of ports.
+
+
+Interacting with your local chain
+=================================
+
+You can now interact with your local chain through the node via the :ref:`Concordium Node gRPC API V2 <grpc2-documentation>` exposed on port 20001, as you would with :ref:`Mainnet<glossary-mainnet>` or :ref:`Testnet<glossary-testnet>`. Concordium provides various :ref:`SDKs and APIs<sdks-apis>` that facilitate this as well as the `Concordium Client <concordium-client>`_ command-line tool. Assuming you have the ``concordium-client`` binary version 5.1.1 or higher in your path, list the accounts using the ``account list`` command:
```
dg-concordium, line 154:
```suggestion
You can now interact with your local chain through the node via the :ref:`Concordium Node gRPC API V2 <grpc2-documentation>` exposed on port 20001 as you would with :ref:`Mainnet<glossary-mainnet>` or :ref:`Testnet<glossary-testnet>`. Concordium provides various :ref:`SDKs and APIs<sdks-apis>` that facilitate this as well as the `Concordium Client <concordium-client>`_ command-line tool. Assuming you have the ``concordium-client`` binary version 5.1.1 or higher in your path, list the accounts using the ``account list`` command:
```

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+.. code-block:: console
+
+    $ concordium-client --grpc-ip 127.0.0.1 --grpc-port 20001  account list
+    Accounts:
+                     Account Address                     Account Names
+    --------------------------------------------------------------------
+    44pozJMswBY5NQdh2MdHLTRQhmZg828wmBCvVckBgsHc7xhiGY
+    4mUMfBFDqFkr3SCQx3k6x8RuWWFyLQHhE2AnJrdk9XtVto8mnK
+
+The two accounts addresses in the output correspond to those of the generated baker and foundation account specified in the ``genesis-creator`` configuration file. You can verify the balance and stake of the baker by supplying the first of the two account addresses to the ``account show`` command:
```
dg-concordium, line 165:
```suggestion
The two accounts' addresses in the output correspond to those of the generated baker and foundation account specified in the ``genesis-creator`` configuration file. You can verify the balance and stake of the baker by supplying the first of the two account addresses to the ``account show`` command:
```

source/mainnet/net/index.rst
dg-concordium: I just built locally and it works fine for me. It shows up in the Develop on Concordium toctree as I would expect and it stays there when I navigate to other pages.

dg-concordium:
I feel like we ought to add a note at the top of this topic that this should only be used in some cases and not attempted by those with little or no blockchain experience. @abizjak ?

source/mainnet/net/index.rst
vikt0r0: That's great :-) I suspected that was the case.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    balance = "1000000000000000"
+    stake = "500000000000000"
+    template = "baker"
+    identityProvider = 0
+    numKeys = 1
+    threshold = 1
+    repeat = 1 # Changed from 5
+
+.. Note::
+
```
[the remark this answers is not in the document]
vikt0r0: It is definitely wrong, but I would like to correct it to let the reader know how to deduce this.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+* the initial protocol version
+* cryptographic parameters
+* anonymity revokers
+* identity providers
+* foundation accounts
+* keys for updating the chain
+* various parameters for the genesis
+
+Furthermore, it specifies where to save the output that is used to invoke the node binary. Many of these options are not relevant when testing smart contracts and the easiest way to get started is to piggyback on one of the examples in the ``./examples`` folder. In the following, you will use the file ``./examples/genesis5.toml`` and modify it slightly. Inspecting the file reveals that it specifies an initial protocol version of 5, to output credentials for 5 baker accounts, 2 foundation accounts, and the genesis time set to the system time at generation. It also specifies 5 seconds as the average time per block. Further inspection of the table at the ``accounts`` key of the file reveals that the bakers each have an initial balance of 10^15 microCCD and a stake of 5 * 10^14 microCCD. Change this section such that only one baker credential is produced by setting the value of the ``repeat`` key to 1 as follows:
```
EmilLa1, line 82: Maybe we should just check in a toml file with a suitable configuration file that the user can simply use.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    stake = "500000000000000"
+    template = "baker"
+    identityProvider = 0
+    numKeys = 1
+    threshold = 1
+    repeat = 1 # Changed from 5
+
+.. Note::
+
+    Note that the staked amount needed to participate in the finalization committee is some fraction of the total amount of existing CCD defined by the value of the ``capitalBound`` key in the configurations file. The total amount is the sum of the balances of all the baker and foundation accounts specified in the genesis configuration file. In this particular example, the stake is sufficient for baking.
```
EmilLa1, line 98: I'm not sure it's important to mention the more fine grained chain parameters. See my comment above.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    cp ./genesis.dat localchain-node-0/
+
+Now run the baker node as follows:
+
+.. code-block:: console
+
+    concordium-node \
+      --no-bootstrap= \
+      --grpc2-listen-addr 127.0.0.1 \
+      --grpc2-listen-port 20001 \
```
EmilLa1, line 139: Not sure that this choice of port in the example is that great. It will conflict with a testnet node (if the user has one running that is).

source/mainnet/net/guides/run-local-chain.rst
EmilLa1: I think it's ok to have. I'm not sure why you are mentioning sets of ports though. You are only specifying one for the grpc. 

Maybe you wanted to include the `listen-port`  for specifying the TCP port that the node uses to receive messages on from other peers.

source/mainnet/net/guides/run-local-chain.rst
vikt0r0: I agree!

source/mainnet/net/guides/run-local-chain.rst
vikt0r0: I mention it since that it would fail if we two node instances attempt to listen on the same network socket. But the phrasing is admittedly incorrect and I will change it :-) thanks!

source/mainnet/net/guides/run-local-chain.rst
vikt0r0: Yes, I think the change itself is not actually necessary, but I think it is nice that the user gets to inspect the file and change a parameter in the guide. But perhaps it is better to omit such details?

source/mainnet/net/guides/run-local-chain.rst
vikt0r0: Yea, let's figure out what to do in the above, and then change this accordingly.

source/mainnet/net/guides/run-local-chain.rst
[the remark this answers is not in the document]
abizjak: I would recommend omitting these details and link to the file instead.

Maybe we should tell them how they can change the balance and number of accounts in that file, but that's it.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+
+.. Note::
+
+    Note that the staked amount needed to participate in the finalization committee is some fraction of the total amount of existing CCD defined by the value of the ``capitalBound`` key in the configurations file. The total amount is the sum of the balances of all the baker and foundation accounts specified in the genesis configuration file. In this particular example, the stake is sufficient for baking.
+
+Next, generate the genesis data:
+
+.. code-block:: console
+
+    $ ./target/release/genesis-creator generate --config ./examples/genesis5.toml
```
abizjak, line 103: And I would use single-baker toml in this example command.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    Baker keys will be generated in ./bakers
+    Cryptographic parameter will be generated in ./global
+    The genesis data will be stored in ./genesis.dat
+    The genesis hash will be written to ./genesis_hash
+    There are 2 accounts in genesis, 1 of which are bakers.
+    Genesis time is set to 2023-05-22 15:08:19.803 UTC.
+    Average block time is set to 5000ms.
+    DONE
+
+In particular, the files of interest to you are ``./genesis.dat``, containing the genesis block data, and ``./bakers/baker-0-credentials.json``, containing the credentials of the single baker account that was created.
```
abizjak, line 118: I think the files in `accounts` are also of interest, and can be imported into concordium-client.

source/mainnet/net/guides/run-local-chain.rst
abizjak: I think the note is useful.

But add that the different nodes must use different config and data dirs. That's a bit implicit now.

abizjak (commented):
I think this is good in scope and the amount details.

But I think we need a bit more details on how you do this using the docker node.

We should provide a docker-compose file. I sent this one to some partners some time ago

```toml
# This is an example configuration for running a local node
​
version: '3'
services:
  mainnet-node:
    container_name: local-node
    image: concordium/mainnet-node:latest
    pull_policy: always
    environment:
      # Where the genesis is located
      - CONCORDIUM_NODE_CONSENSUS_GENESIS_DATA_FILE=/mnt/genesis.dat
​
      # Baker credentials file
      - CONCORDIUM_NODE_BAKER_CREDENTIALS_FILE=/mnt/baker-0-credentials.json
​
      # General node configuration Data and config directories (it's OK if they
      # are the same). This should match the volume mount below. If the location
      # of the mount inside the container is changed, then these should be
      # changed accordingly as well.
      - CONCORDIUM_NODE_DATA_DIR=/mnt/data
      - CONCORDIUM_NODE_CONFIG_DIR=/mnt/data
​
      # port on which the node will listen for incoming connections. This is a
      # port inside the container. It is mapped to an external port by the port
      # mapping in the `ports` section below. If the internal and external ports
      # are going to be different then you should also set
      # `CONCORDIUM_NODE_EXTERNAL_PORT` variable to what the external port value is.
      - CONCORDIUM_NODE_LISTEN_PORT=8888
      # Desired number of nodes to be connected to.
      - CONCORDIUM_NODE_CONNECTION_DESIRED_NODES=0
      # Maximum number of __nodes__ the node will be connected to.
      - CONCORDIUM_NODE_CONNECTION_MAX_ALLOWED_NODES=10
      # Address of the GRPC server
      - CONCORDIUM_NODE_RPC_SERVER_ADDR=0.0.0.0
      # And its port
      - CONCORDIUM_NODE_RPC_SERVER_PORT=10000
      # Address of the V2 GRPC server
      - CONCORDIUM_NODE_GRPC2_LISTEN_PORT=20000
      # And its port
      - CONCORDIUM_NODE_GRPC2_LISTEN_ADDRESS=0.0.0.0
      # maximum number of __connections__ the node can have. This can temporarily be more than
      # the number of peers when incoming connections are processed. This limit
      # ensures that there cannot be too many of those.
      - CONCORDIUM_NODE_CONNECTION_HARD_CONNECTION_LIMIT=20
      # Number of threads to use to process network events. This should be
      # adjusted based on the resources the node has (in combination with
      # `CONCORDIUM_NODE_BAKER_HASKELL_RTS_FLAGS`) below.
      - CONCORDIUM_NODE_CONNECTION_THREAD_POOL_SIZE=2
      # The bootstrapping interval in seconds. This makes the node contact the
      # specified bootstrappers at a given interval to discover new peers.
      - CONCORDIUM_NODE_CONNECTION_BOOTSTRAPPING_INTERVAL=1800
      # Haskell RTS flags to pass to consensus. `-N2` means to use two threads
      # for consensus operations. `-I0` disables the idle garbage collector
      # which reduces CPU load for passive nodes.
      - CONCORDIUM_NODE_BAKER_HASKELL_RTS_FLAGS=-N2,-I0
​
    entrypoint: ["/concordium-node"]
​
    # Exposed ports. The ports the node listens on inside the container (defined
    # by `CONCORDIUM_NODE_LISTEN_PORT` and `CONCORDIUM_NODE_RPC_SERVER_PORT`)
    # should match what is defined here. When running multiple nodes the
    # external ports should be changed so as not to conflict.
    ports:
    - "8888:8888"
    - "10000:10000"
    - "20000:20000"
​
    volumes:
    # The node's database should be stored in a persistent volume so that it
    # survives container restart. In this case we map the **host** directory
    # ~/tmp/local to be used as the node's database directory.
    - ~/tmp/local:/mnt/data:Z
    - ~/tmp/genesis.dat:/mnt/genesis.dat:Z
    - ~/tmp/baker-0-credentials.json:/mnt/baker-0-credentials.json:Z
```

source/mainnet/net/guides/run-local-chain.rst
vikt0r0: Added.

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+    cd genesis-creator
+    git submodule update --init --recursive
+
+The project is then built using ``cargo``:
+
+.. code-block:: console
+
+    cargo build --release
+
+This produces the binary ``./target/release/genesis-creator`` which you run to generate the genesis data.
```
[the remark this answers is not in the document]
vikt0r0: A whole lot better. That's a great command.

source/mainnet/net/guides/run-local-chain.rst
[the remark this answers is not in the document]
vikt0r0: That is better 👍🏼

source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -0,0 +1,185 @@
...
+
+Installing the node distribution
+--------------------------------
+Concordium Node releases exist for Ubuntu, MacOS, Windows, and Docker. See the :ref:`Node Requirements<node-requirements>` section for information on system requirements and detailed instructions on how to obtain, run, and manage a node. To run a baker, you need the ``concordium-node`` binary supplied with the distributions in your path. Upon successful installation of the distribution verify that the binary is in your path:
+
+.. code-block:: console
+
+    $ concordium-node --version
+    concordium_node 5.3.2
+
```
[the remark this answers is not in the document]
vikt0r0: So should I document how to do run a local chain with docker, or are there existing docs that can be linked to?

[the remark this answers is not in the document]
abizjak: I think we should document it here and provide that sample docker-compose file.

If people are experts there is enough information, but I think this guide is meant also for non-experts which we then need to guide by their hand a bit more.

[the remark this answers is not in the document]
vikt0r0: Makes sense. Should we host this file somewhere, should I include it here or can I put it in the `extra` folder?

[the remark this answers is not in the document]
abizjak: We should just include it here inline I think. Like we do with other docker-compose configurations for running mainnet/testnet nodes.

source/mainnet/net/guides/run-local-chain.rst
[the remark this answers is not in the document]
vikt0r0: I left a note about it, but did not specify which parameter. If you want me to I will get rid of it.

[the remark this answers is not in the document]
abizjak: I think it looks good now with the note.

abizjak (approved):

dg-concordium (approved):

PR #872, opened 2023-08-18
Add concordium academy when deploying

limemloh:
## Purpose

Deploy Concordium Academy together with the developer documentation.
This will make the Deploy action also build and deploy Concordium Academy.
Likewise, adding preview label causes a preview of Concordium Academy to be built.

## Changes

- Simplify build script and add Concordium academy.
- Make the 'Deploy' action include deployment of Concordium Academy.
- Generated and added keys to allow for deployment to https://github.com/Concordium/concordium-academy

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -28,7 +28,7 @@ jobs:
         shell: bash
         run: ./scripts/build.sh
 
-      - name: Deploy 🚀
+      - name: Deploy Developer Documentation 🚀
         uses: JamesIves/github-pages-deploy-action@v4.4.1
         with:
           branch: gh-pages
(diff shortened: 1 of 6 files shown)


.github/workflows/deploy.yml
```diff
@@ -28,12 +28,23 @@ jobs:
...
+          clean-exclude: |
+            pr-preview/
+            pr-preview-academy/
+
+      - name: Deploy Concordium Academy 🚀
+        uses: JamesIves/github-pages-deploy-action@v4.4.1
+        with:
+          branch: gh-pages
+          repository-name: Concordium/concordium-academy
+          ssh-key: ${{ secrets.CONCORDIUM_ACADEMY_DEPLOY_KEY }}
```
abizjak, line 48: Where is it documented that this needs to be managed?

Could we have it on Confluence perhaps? Or in a README?

abizjak (approved):

.github/workflows/deploy.yml
limemloh: I have added a section in the Readme of this repo

dg-concordium (approved):

PR #877, opened 2023-09-05
Note for collector when using TLS

EmilLa1:
## Purpose

Explain how to setup the collector if the node is configured with TLS.

**Should not be published before node version 6.1.x.** is released.

## Changes

Added a note.

## Checklist

- [ ] My code follows the style of this project.
- [ ] The code compiles without warnings.
- [ ] I have performed a self-review of the changes.
      hard-to-understand areas.


dg-concordium (approved):
Only question is whether there will be updated config files for the Docker images. We have examples of those in the documentation too, so those may also need to be updated.

EmilLa1:
It should probably also be mentioned in the note box here https://developer.concordium.software/en/mainnet/net/guides/run-node.html talking about the collector.

EmilLa1:
> Only question is whether there will be updated config files for the Docker images. We have examples of those in the documentation too, so those may also need to be updated.

It should probably also be mentioned in the note box here https://developer.concordium.software/en/mainnet/net/guides/run-node.html talking about the collector.

source/mainnet/net/guides/run-node-macos.rst
```diff
@@ -286,6 +286,9 @@ Since the GRPC V2 port is different than the GRPC V1 port, you need to change it
 
 .. _configure-baker-macos:
 
+.. Note::
+  If the node grpc host is configured with TLS, then `CONCORDIUM_NODE_COLLECTOR_GRPC_HOST` must be configured such that it uses the domain of the certificate.
```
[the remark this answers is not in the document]
EmilLa1: Rephrased it a bit so it is consistent with the other pages.

source/mainnet/net/guides/run-node-macos.rst
```diff
@@ -286,6 +286,9 @@ Since the GRPC V2 port is different than the GRPC V1 port, you need to change it
 
 .. _configure-baker-macos:
 
+.. Note::
+   If the node is configured with TLS, then `CONCORDIUM_NODE_COLLECTOR_GRPC_HOST` must be configured such that it uses the domain of the certificate.
```
abizjak, line 290: It would probably make sense to link to the instructions on how to configure TLS, here https://github.com/Concordium/concordium-node/blob/main/docs/grpc2.md#grpc-api-v2

abizjak: And in all the instances below as well.

source/mainnet/net/guides/run-node-ubuntu.rst
```diff
@@ -127,6 +127,9 @@ This will open your overrides in your default editor. Below is an example for th
 
 .. _upgrade-node-Ubuntu:
 
+.. Note::
+   If the node is configured with TLS, then `CONCORDIUM_NODE_COLLECTOR_GRPC_HOST` must be configured such that it uses the domain of the certificate.
```
abizjak, line 131: Please give an example section here, e.g.
```
[Service]
Environment=CONCORDIUM_NODE_COLLECTOR_GRPC_HOST=https://example.concordium-node.io:20000
```

abizjak (approved):

dg-concordium:
Should it also be added here: https://developer.concordium.software/en/mainnet/net/nodes/node-runner-service-configuration.html?

PR #922, opened 2023-11-03
Update schema embedded key

schwartz-concordium:
## Purpose

Update custom section key.


limemloh:
Yes, this is outdated and yes we should cover all 3 possible custom section names used for the schema.

limemloh (approved):

schwartz-concordium:
> Yes, this is outdated and yes we should cover all 3 possible custom section names used for the schema.

I will make a new PR with a section describing how to extract the schema for older modules.

PR #968, opened 2024-01-31
Add tips about using the dapp starter template

limemloh:
## Purpose

This PR adds 'tip' blocks for mentioning the concordium dapp starter template.


dg-concordium (approved):
This should probably also be added to the dApp examples page and the developer resources page.

Bargsteen (approved):

PR #983, opened 2024-03-01
A user guide for the factory pattern.

### LINKED ISSUE #982: Expose Module Reference: Factory pattern guide
Write developer documentation guide for the factory pattern.

td202:
## Purpose

Closes #982 

This adds a user guide on how (and when not) to use the factory pattern for smart contracts.

## Changes

- Add factory pattern guide.

## Checklist

      hard-to-understand areas.


source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+=======================================
+Using the factory pattern on Concordium
+=======================================
+
+.. Warning::
+
+    This guide makes use of features that are first available in protocol version 7,
+    namely getting the contract name and module reference of a smart contract instance
+    from a smart contract.
+    Before the protocol upgrade to version 7, these features will not work.
```
DOBEN, line 12: Could we link to a page where people can check when version 7 will be expected to be live (or give a tentative date here e.g. around summer).

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+may be suitable, depending on the application.
+
+Construct the product directly
+------------------------------
+
+If the factory does not maintain an on-going relationship with the product,
+then it is generally possible to simply initialize the product entirely in
+the constructor (``init_product``), removing the separate ``initialize`` operation entirely.
+In general, if the factory contract does not need to update its state
+after initializing the product, then it serves no purpose. It would be
```
DOBEN, line 115: If I understand correctly, you want to state that only if the `factory` stores some data in its state in the same function call when `producing` a new `product` then it has a purpose. 

Since the factory can also update its state before `initializing` the product which is the recommended pattern `check-effect-interact`, I would rephrase this to:

```suggestion
In general, if the factory contract does not need to update its state
when initializing the product, then it serves no purpose. It would be
```

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+-------------------------
+
+Instead of having separate contract instances for the factory and each product, all of these
+could be combined in a single monolithic contract. In this model, each product is assigned a
+virtual address, and the contract maintains a mapping from virtual addresses to the state of
+each product.
+
+For instance, a CIS2 contract can manage multiple NFTs that each have distinct token IDs.
+Rather than each NFT being its own contract instance (created by a factory), they are simply
+handled as part of the state of the overall CIS2 contract. Here, the token ID actts as a
```
DOBEN, line 140:
```suggestion
handled as part of the state of the overall CIS2 contract. Here, the token ID acts as a
```

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+            .or(Err(FactoryError::NonExistentProduct))?;
+        ensure_eq!(self_module_ref, product_module_ref, FactoryError::InvalidProduct);
+
+If the product were instead defined in a different module from the factory, then the expected
+module reference for the product could be hard-coded into the factory. Alternatively, if the
+module rereference for the product is to be determined later, it could be passed as a parameter
+when creating the ``factory`` instance.
+
+If the module defining the product is known to only contain one smart contract, then checking the
+module reference is sufficient for identifying the code of the product smart contract. In this case,
```
DOBEN, line 211:
```suggestion
module reference is sufficient for identifying the code of the product smart contract. In the example smart contract,
```

People could think that `in this case` refers to the example you describe in the previous sentence.

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+        )
+        .or(Err(FactoryError::InitializeFailed))?;
+
+Here, it is assumed that ``initialize`` will fail, for instance, if it is called on a product that
+has previously been initialized.
+It only remains to update the factory contract's state:
+
+.. code-block:: Rust
+
+        let state = host.state_mut();
```
DOBEN, line 242: The usual smart contract best practice pattern is to do a `check-effect-interact` pattern. This means we should do the state update in the factory before calling an external contract. We can consider to change the order in this function in the factory example.

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+        ensure_eq!(*state, ProductState::Uninitialized, ProductError::AlreadyInitialized);
+
+
+Since the construction and initialization of the product occur in two
+separate transactions, it is possible that a third party might try to hijack
+the process by inserting their own transaction to initialize the product.
+For instance, an adversary could invoke a different factory instance than indended by the user,
```
DOBEN, line 298:
```suggestion
For instance, an adversary could invoke a different factory instance than intended by the user,
```

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+the invoker Adversary does not match the owner User), but success for User:
+
+.. image:: images/factory-adversary2.svg
+    :alt: sequence diagram showing how a hijacking attempt fails
+
+.. Note::
+    Typically, it is wrong to use the invoker of a transaction for
+    authorization, rather than the immediate caller. For instance, a user might
+    invoke some untrusted smart contract, and expect it is not authorized to
+    transfer tokens she holds on another contract. If the token-holding contract
```
DOBEN, line 322:
```suggestion
    transfer tokens it holds on another contract. If the token-holding contract
```

DOBEN (approved):

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+        *state = ProductState::Initialized(product);
+
+
+.. Note::
+
+    If a user invokes ``initialize`` directly (rather than through a factory contract) it will fail.
+    This is because ``initialize`` checks that the immediate caller is a smart contract, and
+    records the contract address in the state. The fact that this prevents a user from directly
+    invoking ``initialize`` is incidental. The design intent is not to prevent the user from
+    badly initializing products (which they could also do by invoking a "BadFactory" as previously
```
dg-concordium, line 365:
```suggestion
    initializing products badly (which they could also do by invoking a "BadFactory" as previously
```

source/mainnet/smart-contracts/guides/factory-pattern.rst
abizjak: We don't have that yet.

source/mainnet/smart-contracts/guides/contract-dev-guides.rst
```diff
@@ -26,3 +26,4 @@ Once you are familiar with smart contracts, it is a good idea to read the :ref:`
    upgradeable-contract
    json-params
    migrate-contracts
+   factory-pattern
```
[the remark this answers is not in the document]
td202: This is not really written as a tutorial, so I'd rather not move it there. I'll add a link from the Ethereum FAQ, but since there aren't other guides under it, I think it's reasonable to keep it in the guides section (for now).

source/mainnet/smart-contracts/guides/factory-pattern.rst
```diff
@@ -0,0 +1,392 @@
...
+
+For instance, a CIS2 contract can manage multiple NFTs that each have distinct token IDs.
+Rather than each NFT being its own contract instance (created by a factory), they are simply
+handled as part of the state of the overall CIS2 contract. Here, the token ID actts as a
+virtual address.
+
+The main disadvantage of this approach is that the isolation between the states of each product
+must be enforced by the contract itself. If the state becomes corrupted (due to a bug in the
+contract) then all of the encapsulated products can potentially be affected. With the factory
+pattern, the runtime system of the blockchain enforces isolation. As always, the balance of risks
```
abizjak, line 146: This argument here about isolation seems spurious because if all contracts are initialized from the same code, then corruption will likely also lead to corruption in all products.

So this is neither here nor there.

abizjak (approved):

PR #992, opened 2024-03-19
Add whitepaper to documentation repo

dg-concordium:
## Purpose

The latest version of the whitepaper should be stored somewhere so that the link stays static for both the documentation site as well as the .com site.

## Changes

Add PDF to _static folder. Tested in glossary.

## Checklist

      hard-to-understand areas.

## CLA acceptance

_Remove if not applicable.


source/mainnet/net/resources/glossary.rst
```diff
@@ -8,7 +7,7 @@
 Glossary of Concordium Terms
 ============================
 
-Also see the Concordium `whitepaper`_ for more details on the terms described below.
+Also see the Concordium `Whitepaper <../../../_static/Concordium-White-Paper-latest.pdf>`_ for more details on the terms described below.
```
limemloh, line 10: **Issue:** The file is not linked correctly in the production build.

The file is not located in the same relative path in the production build. The sphinx way to link to non-rst documents like this, is using the [`:download:` role](https://www.sphinx-doc.org/en/master/usage/referencing.html#role-download)
```suggestion
Also see the Concordium :download:`Whitepaper <../../../_static/Concordium-White-Paper-latest.pdf>` for more details on the terms described below.
```
This will work both locally and in production, and give warnings if the file does not exist. One downside seems to be that it is styled differently.

dg-concordium: You mean the PDF is styled differently? Because I took a really old one...so maybe the style has changed since then.

dg-concordium: Also, I don't like the styling that it puts in front of the download role...with the little download arrow.

dg-concordium: Let me see what i can do in the CSS to make the styling prettier.

limemloh: It was the download arrow I meant. Just the link to the file is styled differently, the PDF is the same.

chportma (approved):

PR #996, opened 2024-03-21
adding Election Rules 2024 and GC Rules to the public documents

chportma:
## Purpose

Making new governance docs public

## Changes

Adding two files

public/governance/rules/The Governance Committee.pdf
public/governance/rules/ElectionRules2024.pdf

They have been reviewed and approved by the GC and EMG.


tschudid (approved):
Looks good to me

PR #1016, opened 2024-05-16
whitepaper without shielded and anonymous transactions

chportma:
Whitepaper updated to remove all mention of shielded and anonymous transactions. Only the file public/governance/whitepaper/Concordium White Paper.pdf has changed and is now v.1.9.


tschudid (approved):
Looks good to me

PR #1019, opened 2024-05-22
Remove anon word

DOBEN:
## Purpose

Update the `developer documentation`:

The word `anonymity` made it difficult to progress in important talks. The decision was made to clarify/re-phrase these areas across our documentation.

## Changes

- Rename `anonymity revokers` to `identity disclosure authority`.
- Rename `anonymity revoker threshold` to `identity disclosure threshold`.
- Rename `Anonymity revocation` to `disclosing an identity`.


source/mainnet/net/guides/run-local-chain.rst
```diff
@@ -71,7 +71,7 @@ The ``genesis-creator`` tool uses a TOML configuration file format for specifyin
 
 * the initial protocol version
 * cryptographic parameters
-* anonymity revokers
+* anonymity revokers (Identity disclosing authorities)
```
erasmus, line 74: Can we replace with `identity disclosing authorities` only?

erasmus (commented):
In general, I think it makes sense to keep nouns and verbs consistent.

So, I propose:

* Identity Disclosure Authority (start cased in headlines, lower case in body text)
* identity disclosure threshold
* disclosing an identity
* undisclosed identity, etc. 

If we use both disclose and reveal that might lead to some confusion

source/mainnet/net/guides/run-local-chain.rst
DOBEN: We would need to update the ``genesis-creator`` tool to use a different TOML file configuration. 
```
The ``genesis-creator`` tool uses a TOML configuration file format for specifying parameters from which the genesis data is generated. The TOML file specifies:
...
* anonymity revokers (Identity disclosing authorities)

```

erasmus (approved):

PR #1022, opened 2024-05-28
Add guide on multi-sig feature in concordium-client

### LINKED ISSUE #1017: `Multi-sig` feature documentation of `concordium-client`
**Task description**

Document `multi-sig` feature from this PR:
https://github.com/Concordium/concordium-client/pull/300

Add guides on how to create a multi-sig account on Concordium (two flavors):
- An account with several keys (per identity).
- An account with several identities.
The guides should consolidate the knowledge from these two guides:
https://gist.github.com/DOBEN/683fe1a7c82a0551546a7ec242d30cc0
https://gist.github.com/limemloh/8c0c55f67cf5a83ac7cc21cb646e65c1

**Sub-tasks**
- [x] Guide on how to create a multi-sig account.
- [x] Guide on how to configure `concordium-client` to have access to the keys (of a multi-sig account).
- [x] Guide/Tutorial/Example flow of a multi-sig transaction executed with `concordium-client`.

DOBEN:
## Purpose

closes #1017 

## Changes

Add multi-sig feature in ``concordium-client`` tool


source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+         "keys": {
+            "0": {
+               "verifyKey": "<Hex encoding without '0x' of the public key already on-chain>"
+            },
+            "1": {
+               "verifyKey": "<Hex encoding without '0x' of the public key from your newly generated key pair>"
+            }
+         },
+         "threshold": 2
+      }
```
soerenbf, line 123: Not a comment to the documentation, but would it make sense to have an easier way to do this? Something like

```sh
concordium-client account add-keys \
  --credId <credential-registration-id> \
  --sender <account-name-or-address> \
  --key <verify-key> # --key can have multiple values
```

Just a random thought

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+   in the future and to replace the desktop wallet's file format.
+
+To add additional keys to an existing account on Concordium using the ``concordium-client`` tool, you can either:
+
+.. dropdown:: Add an additional public-private key pair to your current credential
+
+   - Step 1: Create an additional Ed25519 public-private key pair as you feel safe.
+
+   For testing purposes we recommend, the `front-end <https://cyphr.me/ed25519_tool/ed.html>`_ to generate an un-safe
+   additional key pair.
```
soerenbf, line 60: Would it make sense to recommend a safe way to do this is as well, or do we want to avoid that for legal reasons?

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+   The above file specifies all keys that are associated with your account for a given credential after the update and
+   the threshold can be set accordingly to a number between 1 and the number of keys.
+
+.. dropdown:: Add an additional credential with at least one public-private key pair
+
+   - Step 1: Create an additional credential.
+
+   We recommend using the flow in the desktop wallet to :ref:`create a credential file<create-credentials-file>`.
+   Alternatively, you can `clone the project <https://github.com/Concordium/concordium-base/blob/main/rust-bins/src/bin/user_cli.rs>`_
+   and run its command locally to generate a credential file.
```
soerenbf, line 140: Do we plan to add support to generate a credentials file in concordium client?

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+               "threshold": 1
+            },
+            "credentials": {
+               "0": "97f325c9f86066ab0c80ff879c21629eb67818841940869308d6a72886d18f8668e62e43ad228fdcbda245d0722454df"
+            },
+            "address": "4jxvYasaPncfmCFCLZCvuL5cZuvR5HAQezCHZH7ZA7AGsRYpix"
+         }
+      }
+
+   You can import the keys of this ``browser_wallet.export`` file and associate it to your account as follows:
```
soerenbf, line 222: In the dropdown title here it states that you should "adapt" the browser wallet export. I don't see anything in the section regarding adapting the exported file?

Maybe a comment in the JSON file stating which section was added here, or some other way for users to easily identify what was added to the export file?

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+
+   If you already imported the new key to ``concordium-client`` but it is associated with a different account
+   (e.g. because the public-private key pair was generated by creating a new account on Concordium which was imported
+   to ``concordium-client`` and this key pair was re-used in your multi-sig account),
+   you can look up the key in the local key directory of the ``concordium-client`` tool and associate
+   the file's content to your multi-sig account.
+
+   .. note::
+
+      We do not recommend to re-use keys on different accounts in production.
```
soerenbf, line 238: What _do_ we recommend then? I feel like if we don't provide a recommended path for the user to follow, they will feel stuck.

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+                  },
+                  "cipherText": ...
+               },
+               "verifyKey": ...,
+               "schemeId": "Ed25519"
+            },
+            ...
+         },
+         ...
+      }
```
soerenbf, line 268: How should the user get the values for this file? I feel like this would be confusing to follow (at least I don't understand what you're supposed to do to get to this point).

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+            },
+            "address": "4jxvYasaPncfmCFCLZCvuL5cZuvR5HAQezCHZH7ZA7AGsRYpix"
+         }
+      }
+
+   You can import the keys of this ``browser_wallet.export`` file and associate it to your account as follows:
+
+   .. code-block:: console
+
+      $concordium-client config account import browser_wallet.export --name <choose-a-name-for-your-account>
```
soerenbf, line 226: Maybe wrap the section up explaining what was achieved with the sequence of steps

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+            ...
+         },
+         ...
+      }
+
+.. note::
+
+   These commands update the keys in the key directory of your local ``concordium-client`` tool. No transaction
+   is sent on-chain.
+
```
soerenbf, line 274: And also here it would be nice to state what has been achieved

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+         "transactionType": "transfer"
+      },
+      "signature": {
+         "0": {
+            "0": "3099534c5f32daf64dc40b7a0013979b9b74b167d259fc787a363ed2db7f1bcdafdcf06e166b2d915c7f29043186b3015a6064755bf3c3733bca2151b2b19c04"
+         }
+      },
+      "signer": "4jxvYasaPncfmCFCLZCvuL5cZuvR5HAQezCHZH7ZA7AGsRYpix",
+      "version": 1
+   }
```
soerenbf, line 321: Should state somewhere that the transaction is created for a basic account with only one set of keys. I was kind of expecting multiple signatures as the paragraph just before stated "using all local keys" combined with the fact that we're describing multi-sig accounts/transactions

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+``UpdateCredential``, and ``UpdateCredentialsKeys``.
+
+To create a multi-sig transaction, execute a transaction-creating command and add an ``--out`` flag.
+This flag will cause ``concordium-client`` to output the multi-sig
+transaction into the specified file instead of sending it on-chain.
+
+If you omit the ``--signers`` flag, ``concordium-client`` will output
+a transaction signed with all keys associated with the account as present in the local key directory.
+If you want to sign with specific keys from the local key directory,
+you can specify some of them with the ``--signers`` flag (e.g. ``--signer "0:0,0:1"``).
```
soerenbf, line 291: I assume theres a typo here?

```suggestion
you can specify some of them with the ``--signers`` flag (e.g. ``--signers "0:0,0:1"``).
```

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,448 @@
...
+
+.. code-block:: json
+   :force:
+
+   {}
+
+.. note::
+
+   Choose an ``expiryTime`` for the transaction that takes into account the time it takes to gather all signatures
+   by the different entities. If the ``expiryTime`` has passed, the transaction is no longer valid.
```
soerenbf, line 374: Should we not provide examples of how to choose an expiry time? Is this possible to do in the `transaction send` command? If so, we should add it there.

soerenbf (commented):
Here are a few comments

source/mainnet/net/references/multi-sig.rst
DOBEN: I updated the text and gave a better example.

source/mainnet/net/references/multi-sig.rst
soerenbf: I think openSSH `ssh-keygen` would be a safe bet.

```sh
ssh-keygen -t ed25519
```

This is at least the approach recommended by github and others for generating ssh keys. Do you think this would work?

source/mainnet/net/references/multi-sig.rst
```diff
@@ -0,0 +1,541 @@
...
+.. dropdown:: Import the keys by adapting the key export file format from the browser wallet
+
+   We recommend that you export a key file from a random account on the browser wallet and use this
+   JSON file as a template to generate a corresponding JSON file for the keys of the account you want to update.
+
+   .. note::
+
+      You can read up on how to export a key file from the browser wallet :ref:`here<export-key>`.
+
+   Your ``browser_wallet.export`` file should look similar to the following browser wallet export template which has
```
soerenbf, line 209: Would explicitly state that we're talking about the adaptation here

```suggestion
   Your adapted ``browser_wallet.export`` file should look similar to the following browser wallet export template which has
```

source/mainnet/net/references/multi-sig.rst
soerenbf: I still don't think it's entirely clear how the user will actually get the values. It would be helpful to explain where the directory is on the host machine. Or is it possible to print these values somehow?

soerenbf (approved):

source/mainnet/net/references/multi-sig.rst
soerenbf: While that is true, I think in some cases it could be useful to do anyway, as it would be easier for the user to identify exactly what is being added. Also, there actually is a JSON format with comments (`jsonc`), but that is probably not accepted by `concordium-client`.

source/mainnet/net/references/multi-sig.rst
DOBEN: Added it.

PR #1023, opened 2024-05-29
new whitepaper without anon

chportma:
Update whitepaper to version 1.9.1: does not contain anon* any longer


tschudid (approved):
Looks good to me

PR #1024, opened 2024-06-03
Added Concordium Client voting instructions

chportma:
## Purpose

The GC voting documentation needs instructions for Concordium Client users

## Changes

- Added instructions for Concordium Client users for delegating a vote.
- Removed recommendations to delegate to the mobile wallet. Recommendations are to delegate to the browser wallet and CryptoX wallet.


chportma:
@mh-concordium , please check that the command for the concordium client to delegate a vote and the explanation around it are correct. Thanks!

source/mainnet/net/voting/voting.rst
```diff
@@ -118,6 +120,20 @@ These older wallets do not have the capability to connect to dApps. But you do h
...
 
+.. dropdown:: Concordium Client
+
+    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Concordium Client wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before the election.
+
+    **To delegate your vote:**
+
+    .. code-block:: console
+
+        $concordium-client transaction send --amount AMOUNT --receiver A --sender B --memo delegatevote2024
```
mh-concordium, line 131: I have checked the command and indeed I receive tx with the correct memo. However, notice that this command works only if user has a local node running.

For better usability, we would add --grpc-ip grpc.mainnet.concordium.software so that they can just use grpc node. Also, on that node we have TLS so --secure flag needs to be used. So:
`concordium-client --secure --grpc-ip grpc.mainnet.concordium.software transaction send --amount AMOUNT --receiver A --sender B --memo delegatevote2024`

source/mainnet/net/voting/voting.rst
```diff
@@ -118,6 +120,20 @@ These older wallets do not have the capability to connect to dApps. But you do h
...
+
+    #. Enter the command above in the Concordium Client, where AMOUNT is the number of CCD to be sent (1 micro-CCD is enough), A is the name/address of the account to which the vote is delegated, and B is an optional sender name/address, which is needed only if multiple accounts have been imported in the Concordium Client. Note the memo *delegatevote2024* for the delegation to be valid.
```
mh-concordium, line 133:
```suggestion
    #. Enter the command above in the Concordium Client, where AMOUNT is the number of CCD to be sent (1 micro-CCD is enough), A is the name/address of the account to which the vote is delegated, and B is sender name/address. Note the memo *delegatevote2024* for the delegation to be valid.
```

In my opinion, it would be better if they always try to add the name or address to be sure it is clear what is going to happen.

source/mainnet/net/voting/voting.rst
```diff
@@ -52,20 +54,20 @@ Since the votes are weighted by the number of CCD on the account, for a user to
...
+==================================================================
 
 These older wallets do not have the capability to connect to dApps. But you do have an option to exercise your vote.
 
 .. dropdown:: Desktop wallet
 
-    Users need to create a new account in |bw|, |cryptox|, or |mw-gen2| and delegate their vote from their Desktop wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw|, |cryptox|, or |mw-gen2| at any time before the election.
+    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Desktop wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before delegating your vote.
 
     **To delegate your vote:**
```
mh-concordium, line 66: The way I read it, it is not explicitly said that delegation transaction must be sent while the voting window time is open. Otherwise, delegation will not be picked up.

source/mainnet/net/voting/voting.rst
```diff
@@ -118,6 +120,20 @@ These older wallets do not have the capability to connect to dApps. But you do h
...
+.. dropdown:: Concordium Client
+
+    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Concordium Client wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before the election.
+
+    **To delegate your vote:**
+
+    .. code-block:: console
+
+        $concordium-client transaction send --amount AMOUNT --receiver A --sender B --memo delegatevote2024
+
```
mh-concordium, line 132: In regards to the grpc that will be used, in agreement with SRE and Søren we should add disclaimer:
_There is no service license agreement for the grpc endpoint._

mh-concordium (changes requested):
Looks good. I had some comments and asked for changes.

soerenbf (approved):
I agree with martins comments, otherwise it looks good to me too! 😄

source/mainnet/net/voting/voting.rst
chportma: It isn't clear to me exactly where I should put this disclaimer. I will integrated it in the numbered list item describing the concordium client command. If you prefer something else, please make another suggestion.

source/mainnet/net/voting/voting.rst
```diff
@@ -61,7 +61,7 @@ These older wallets do not have the capability to connect to dApps. But you do h
 
 .. dropdown:: Desktop wallet
 
-    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Desktop wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before delegating your vote.
+    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Desktop wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before delegating your vote. But the memo transaction which effectively delegates the vote and the vote from the new account need to take place within the official voting window.
```
soerenbf, line 64: This sentence read a little weird for me. I don't know if theres another form of punctuation more suited for this? 🤷 

```suggestion
    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Desktop wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before delegating your vote. But the memo transaction, which effectively delegates the vote, and the vote from the new account both need to take place within the official voting window.
```

source/mainnet/net/voting/voting.rst
```diff
@@ -96,7 +96,7 @@ These older wallets do not have the capability to connect to dApps. But you do h
 
    **Alternative route**
 
-   Alternatively, users need to create a new account in |bw| or |cryptox|, and delegate their vote from their |mw-gen1| account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before delegating your vote.
+   Alternatively, users need to create a new account in |bw| or |cryptox|, and delegate their vote from their |mw-gen1| account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before delegating your vote. But the memo transaction which effectively delegates the vote and the vote from the new account need to take place within the official voting window.
```
soerenbf, line 99: And here

source/mainnet/net/voting/voting.rst
```diff
@@ -122,15 +122,15 @@ These older wallets do not have the capability to connect to dApps. But you do h
 
 .. dropdown:: Concordium Client
 
-    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Concordium Client wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before the election.
+    Users need to create a new account in |bw| or |cryptox| and delegate their vote from their Concordium Client wallet account(s) to the new account, then vote from the new account. You can create the new account in |bw| or |cryptox| at any time before the election. But the memo transaction which effectively delegates the vote and the vote from the new account need to take place within the official voting window.
```
soerenbf, line 125: And here

soerenbf (approved):

source/mainnet/net/voting/voting.rst
mh-concordium: It's fine. Thank you.

mh-concordium (approved):
Looks good.

PR #1043, opened 2024-07-24
Streamline counter, wCCD and voting smart contract tutorials

Victor-N-Suadicani:
## Purpose

Streamline and update the tutorials to make them simpler and easier to follow.

## Changes

- Rewrote counter tutorial to improve the wording and simplify the code.
  - Many subpar VSCode screenshots that felt unnecessary were removed.
  - The code was simplified to use a single `update` function rather than separate `increment` and `decrement` functions. A single error variant `OverflowError` is used instead of separate errors for the two previous functions.
  - Added instructions for how to use Concordium's testnet node to go through the tutorial, which makes it much easier to follow.
- Slightly edited the wCCD tutorial, including merging the last part with the previous part since it is so short.
- Some cleaning up for the voting tutorial, including making it adhere to the new simplified version, see https://github.com/Concordium/concordium-rust-smart-contracts/pull/440.
- Removed some redundant/repeated sections in the piggybank tutorial and elsewhere.

## Checklist

      hard-to-understand areas.
- [x] (If necessary) I have updated the CHANGELOG.


source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
```diff
@@ -2,29 +2,31 @@
...
-the front end, users will also need to download a web wallet as a browser extension.
-The installation and safe usage of the |bw| might be new for people using your front end.
+You can lower the bar for entry by coding an appealing frontend that provides additional information
+to your users. Web frontends are a familiar sight nowadays, but to use
+the frontend, users will also need to download the |bw| (or simply "the browser wallet") as a browser extension.
+The installation and safe usage of the browser wallet might be new for people using your frontend.
+
 Providing comprehensive explanations and step-by-step guides on your website on topics
-related to the |bw| is important for a good user experience. The |bw|
+related to the browser wallet is important for a good user experience. The browser wallet
```
DOBEN, line 21: I agree with you that the naming `browser wallet` is much more natural than `|bw| = Wallet for Web` and I did not support the naming decision in the past. We need to get a consensus in the company to rename it throughout the documentation since in the past people agreed and pushed for calling it `Wallet for Web`.

As such, we should change the |bw| to `browser wallet` (if agreed internally) instead of partially using the constants and plain `browser wallet` in the text. The current mix will make it hard to have consistent naming throughout the documentation.

source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
```diff
@@ -37,116 +39,109 @@ need for the user to host their own Concordium node.
 |bw|
 -------------------------
 
-A web wallet is a piece of code that can be added as an extension to supported browsers such as ``Chrome``.
-The web wallet allows you to interact with the chain and make transactions.
-Currently, the |bw| does this by connecting to a (gRPC) server that communicates with a node.
-The |bw| hosts the private keys corresponding to the accounts of the user.
+Concordium's browser wallet is an extension to Chromium-based browsers such as Google Chrome.
```
DOBEN, line 42: everywhere in the documentation:

```suggestion
Concordium's |bw| is an extension to Chromium-based browsers such as Google Chrome.
```

source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
```diff
@@ -37,116 +39,109 @@ need for the user to host their own Concordium node.
...
+and sends it to the browser wallet extension. The transaction object is signed by the private key hosted in the wallet
+and transmitted to the server via HTTPS. This server has access to a Concordium node and converts
+the request (including the signed transaction object) that comes via HTTPS from the browser wallet
 to a request that the Concordium node can execute. The signed transaction is
 transmitted via peer-to-peer communication to other Concordium nodes and becomes
 part of the Concordium blockchain.
 
 Installation
-^^^^^^^^^^^^
+-------------------------
```
DOBEN, line 55: Should have the same length as the heading
```suggestion
--------------
```

source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
```diff
@@ -37,116 +39,109 @@ need for the user to host their own Concordium node.
...
+.. image:: ../wCCD/images/wCCD_tutorial_21.png
+    :width: 30 %
 
-    .. image:: ../wCCD/images/wCCD_tutorial_22.png
-        :width: 30 %
+.. image:: ../wCCD/images/wCCD_tutorial_22.png
+    :width: 30 %
 
-    You are ready now to create a new account on testnet.
+Finally, make a new account on testnet.
```
DOBEN, line 107:
```suggestion
Finally, create a new account on testnet.
```

source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
```diff
@@ -37,116 +39,109 @@ need for the user to host their own Concordium node.
...
 
-    .. image:: ../wCCD/images/wCCD_tutorial_20.png
-        :width: 30 %
+.. image:: ../wCCD/images/wCCD_tutorial_20.png
+    :width: 30 %
 
-    You completed the |bw| setup. Send some CCD to your new account or request some CCD from the :ref:`testnet faucet button<testnet-faucet>` within the |bw|.
-    Check that your account balance is displayed and you have enough
-    CCD to be able to execute transactions.
+For testing you'll need some CCD in your account. Send some CCD to your new account or request some CCD from the :ref:`testnet faucet button<testnet-faucet>`.
```
DOBEN, line 115: Testnet faucets are usually websites, but our faucet (where the link points to) is built in the wallet itself (this is rather unusual) and as such good to mention since people expect a website to pop up.

```suggestion
For testing you'll need some CCD in your account. Send some CCD to your new account or request some CCD from the :ref:`testnet faucet button<testnet-faucet>` within the |bw|.
```

source/mainnet/smart-contracts/tutorials/voting/voting-sc.rst
```diff
@@ -23,48 +23,44 @@ Before you start, make sure to have the necessary tooling to build Rust contract
...
 
 The source code of your smart contract is going to be in the ``src`` directory, which already contains the file ``lib.rs``, assuming you follow the above guide
 to set up your project.
 
-Open ``src/lib.rs`` in your editor and you'll see some code for :ref:`writing tests<piggy-bank-testing>`,
+The smart contract template also includes some examples tests under the ``tests`` directory,
 which you can delete for now. You will come back to tests later in this tutorial.
 
-First, bring everything from the |concordium-std|_ library into scope by adding the line:
+In ``lib.rs``, start by bringing everything from the |concordium-std|_ library into scope by adding the line:
```
DOBEN, line 35:
```suggestion
In the ``lib.rs`` file, start by bringing everything from the |concordium-std|_ library into scope by adding the line:
```

source/mainnet/smart-contracts/tutorials/voting/voting-sc.rst
```diff
@@ -86,4 +85,20 @@ And if the end time has passed, an error occurs.
...
 ------------
 
-The election is open from the point in time that this smart contract is initialized until the `end_time`.
+The election is open from the point in time that this smart contract is initialized until the deadline.
+
+Performance considerations
+--------------------------
+
+An important aspect of the voting smart contract to highlight is the way that votes are tallied. Note how two ``HashMap``s are used: one for which accounts have voted for which option, and one for how many votes each option has.
+However, if you think about it, only the first ``HashMap``, the ``ballots`` field is necessary. The vote counts can always be determined from those. So why do it in this way?
```
DOBEN, line 94: The cursive styling does not stop after ``HashMap``.

source/mainnet/smart-contracts/tutorials/voting/voting-sc.rst
```diff
@@ -86,4 +85,20 @@ And if the end time has passed, an error occurs.
 Initializing
 ------------
 
-The election is open from the point in time that this smart contract is initialized until the `end_time`.
+The election is open from the point in time that this smart contract is initialized until the deadline.
+
+Performance considerations
+--------------------------
+
+An important aspect of the voting smart contract to highlight is the way that votes are tallied. Note how two ``HashMap``s are used: one for which accounts have voted for which option, and one for how many votes each option has.
```
DOBEN, line 93: Not sure if that is better:

```suggestion
An important aspect of the voting smart contract to highlight is the way that votes are tallied. Note how two ``HashMap``s are used: one that maps accounts to the option they voted for, and one that maps the voting option to the number of votes it received.
```

source/mainnet/smart-contracts/tutorials/voting/voting-sc.rst
```diff
@@ -86,4 +85,20 @@ And if the end time has passed, an error occurs.
...
+
+The answer is that by counting the votes separately, we avoid having to iterate through all the ballots in order to count the votes.
+Unbounded iterations in smart contracts is in general a bad idea. For instance, if there are many votes, executing the smart contract might become costly, as iterating through the ballots takes many operations.
+In the worst case, this may exceed the energy budget of running the functions, which could make the smart contract unusable.
+
+It's important to keep these performance considerations in mind when writing smart contracts. In general, writing smart contracts requires a different methodology than most other usual software development.
+Be sure to think carefully and read up on the common issues and security problems that may occur when you develop smart contracts.
+
+
+In the next part of the tutorial, we will set up a frontend to make it easier to interact with the smart contract.
```
DOBEN, line 104: A few things to improve/consider:

- Rhetoric questions might not be needed, you can explain the reason for the architectural design in a concise manner without it.

---------------------------------------------
Your description in the smart contract seems quite clear, I suggest re-using it here:
```
    /// Note the data duplication here! The tally can of course always be
    /// determined by examining the ballots field. This requires looping
    /// over all ballots, counting how many votes exist for a specific option.
    /// However, such a loop would be practically unbounded, as it is only
    /// limited by the number of votes. Such a loop could exhaust the energy
    /// budget of the smart contract transaction, potentially making the smart
    /// contract unusable and vulnerable to a kind of DDoS attack.
    ///
    /// Thus, we favor duplicating the data in another hashmap, where the vote
    /// count can be retrieved and updated in constant time. 
```



In the worst case, this may exceed the energy budget of the ``transaction``, which could make the smart contract unusable.


People could ask why not just increase the `energy budget of the transaction` (this is a value set by users and can be set to a very high value).

There are limits either by e.g. the block gas/energy limit or by the limit/funds that the account has at its disposal (you can spend in gas/energy only as much as you have funds on the account). Not sure how to communicate this concisely so.

source/mainnet/smart-contracts/tutorials/wCCD/wCCD-interacting.rst
```diff
@@ -155,12 +155,6 @@ You are ready now to invoke the ``balanceOf`` function with the following comman
 .. image:: ./images/wCCD_tutorial_4.png
     :width: 100 %
 
-.. note::
```
DOBEN, line 158: Why remove this information?

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
```diff
@@ -25,215 +25,211 @@ Once you have set up the tools needed you are ready to create your smart contrac
...
-    #[derive(Debug, PartialEq, Eq, Reject, Serial, SchemaType)]
-    enum Error {
+    /// Errors that may be emitted by this smart contract.
+    #[derive(Debug, PartialEq, Eq, Reject, Serialize, SchemaType)]
+    pub enum Error {
         /// Failed parsing the parameter.
         #[from(ParseError)]
-        ParseParamsError,
-        /// Your error
+        ParseParams,
```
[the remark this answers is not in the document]
Victor-N-Suadicani: Hmmm actually I just made this align to the default template, since that one uses `ParseParams`, so this is more like a template change than a change here in the tutorial.

Excluding the `Error` suffix is actually correct style. For instance, see this [Clippy lint](https://rust-lang.github.io/rust-clippy/master/index.html#/enum_variant_names). Generally enums should not repeat their names in the variants. Then you avoid these repetitions like `Error::SomeError`, and instead just have `Error::Some`. But of course the error should not necessarily be displayed like that when printed (i.e. its `Display` implementation probably ought to show `Error::SomeError` as `"Some error occurred" or something like that). It sounds like the real problem is that these other tools use the `Debug` representation of the error when really they shouldn't?

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
```diff
@@ -25,215 +25,211 @@ Once you have set up the tools needed you are ready to create your smart contrac
...
-        ensure!(param > 0, Error::IncrementError);
-        state.counter += param;
-        Ok(())
-    }
-
-Decrement counter
------------------
-
-Add a new mutable function to implement decrement with a similar approach. It will also take an input parameter, but this time make sure that it is negative because a violation will be caused by an ``Error::DecrementError``. Like the other one, this can be triggered by only the owner of the contract,otherwise it will throw an ``Error::OwnerError``.
+        // Returns Error::ParseError on failure.
```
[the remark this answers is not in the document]
Victor-N-Suadicani: See other comment, this comes from the default template and `Error::ParseError` is actually bad style.

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
```diff
@@ -25,215 +25,211 @@ Once you have set up the tools needed you are ready to create your smart contrac
...
 
-.. image:: images/select-default.png
-    :width: 100%
+The result is a basic skeleton of a smart contract. Initially, it has a ``State`` struct, an ``init`` function for creating new instances, an ``Error`` enum for custom errors, a ``view`` function to read the state, and a dummy ``receive`` function.
 
-Then it will ask for a name and a description of your project. Fill them in.
-The result is a basic skeleton of a smart contract.
-Initially, it has a ``State`` struct, an ``init`` function for creating new instances, an ``Error`` enum for custom errors, a ``view`` function, and a ``receive`` function.
-
-.. image:: images/contract.png
```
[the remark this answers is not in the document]
Victor-N-Suadicani: I disagree - I don't think the screenshots are very helpful, even for beginners. Even if they help a little, they are much more likely to get out of date than the rest of the tutorial, so I'd rather not include them, as they'll likely soon be more confusing than helpful.

>developer audience who can barely code

Is this reasonable? I don't think it's realistic to expect anyone "who can barely code" to get through these tutorials with much understanding. Have we actually tested the tutorial on such people? If you ask me, this tutorial already assumes plenty of stuff that would make it totally infeasible for anyone on a lower level than a hobby developer.

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
```diff
@@ -25,215 +25,211 @@ Once you have set up the tools needed you are ready to create your smart contrac
...
 
-Deploy it with the command below.
+Or, if you just want to try things out on testnet, you can use the testing nodes provided by Concordium:
 
 .. code-block:: console
 
-    $concordium-client module deploy dist/module.wasm.v1 --sender <YOUR-ACCOUNT> --name counter --grpc-port 20001
+    $concordium-client module deploy dist/module.wasm.v1 \
+        --sender <YOUR-TESTNET-ADDRESS> \
+        --name counter \
```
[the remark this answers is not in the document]
Victor-N-Suadicani: I'm not sure what you're saying, do you mean naming the `module.wasm.v1` file? Or what are we setting a name for so we don't need the name parameter here?

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
```diff
@@ -25,215 +25,211 @@ Once you have set up the tools needed you are ready to create your smart contrac
...
 .. code-block:: console
 
-    $concordium-client contract update <YOUR-CONTRACT-INSTANCE> --entrypoint increment --parameter-json <PATH-TO-JSON> --schema dist/smart-contract-multi/schema.bin --sender <YOUR-ADDRESS> --energy 6000 --grpc-port 20001
-
-Start by testing with your conditions. First, try another account other than the owner of the contract since you want that only the owner can call this function.
+    echo 42 > input.json
 
-.. image:: images/owner-error.png
-    :width: 100%
+Now we can invoke the update function with that input by using a contract update transaction. From the point of this transaction in the blockchain, your smart contract will have its new value. If you have your own node, you can invoke the update function like so:
```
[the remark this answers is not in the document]
Victor-N-Suadicani: I guess I'm not using the terminology correctly or maybe not understanding completely.

I mean that by invoking the update function, some kind of transaction or something will be written on the blockchain. Is that not correct? I.e. some kind of transaction will say "this function was invoked with these parameters" and from that point forward in the blockchain, the state will be mutated according to what the function did.

Do you understand what I mean? Maybe you could suggest a better wording?

source/mainnet/smart-contracts/tutorials/piggy-bank/index.rst
```diff
@@ -32,9 +32,11 @@ Finally, the :ref:`fourth part<piggy-bank-frontend>` explains how to set up a we
 .. warning::
 
    The reader is assumed to have basic knowledge of what a blockchain and smart
-   contracts are, and to have some experience with Rust_.
+   contracts are, and to have some experience with Rust_. Consider reading the
+   simpler :ref:`Counter smart contract tutorial <counter-sc>` first.
 
 Before starting the tutorial, make sure that you have all of the necessary tools installed and running as described in :ref:`Setup the development environment<setup-env>`.
+Also, make sure to have a text editor for writing Rust.
```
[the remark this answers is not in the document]
Victor-N-Suadicani: I agree, I mostly added it to be consistent with elsewhere, but I'll happily remove it from those other places too tbh. It does feel very obvious.

source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
Victor-N-Suadicani: My biggest complaint is that `|bw|` expands to `Concordium Wallet for Web` which is just _super long_ and really only makes sense when referring to the wallet in a title or maybe the first time you mention it. I think having `|bw|` expand to that is okay, but it makes no sense to use that expansion everywhere in the prose, as it makes the writing very unnatural.

I don't even know if using such expansions provide a big benefit, finding stuff by a simple search or regex is fine I think? But yea we should agree on how we refer to these things.

Victor-N-Suadicani: To be clear, I am not advocating for changing the official name of the wallet or anything. I am using "browser wallet" like a pronoun or a shorthand to refer to "Concordium Wallet for Web".

I personally find that "web wallet" is a bit of an inaccurate name though, as it is not a web application, but a browser extension.

source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
Victor-N-Suadicani: This will expand to "Concordium's Concordium Wallet for Web", but even fixing the double Concordium, I think the long title just sounds super unnatural in a casual sentence like this.

source/mainnet/smart-contracts/tutorials/voting/voting-dapp.rst
Victor-N-Suadicani: Quite the nitpick but fine :sweat_smile:

source/mainnet/smart-contracts/tutorials/voting/voting-sc.rst
Victor-N-Suadicani: Huh, I guess it doesn't like the s directly after the code/monospace (not cursive) styling? Will see if I can fix that.

source/mainnet/smart-contracts/tutorials/wCCD/wCCD-interacting.rst
Victor-N-Suadicani: The exact same note is repeated at the start of this part of the tutorial, line 71. It felt weirdly repetitive when I read it.

source/mainnet/smart-contracts/tutorials/voting/voting-sc.rst
Victor-N-Suadicani: I reworded/used what I had already written in the code :)

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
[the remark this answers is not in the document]
DOBEN: You describe the recommended way of naming the Error enums in the Rust world. The point I am trying to make is that there is a custom error reporting mechanism in the Concordium smart contracts (using error schemas). This custom error reporting tries to facilitate reporting of errors when there is a chain of contract calls that reject/revert. Calling the variant "ParseParamsError" seems to make sense in the given Concordium smart contract environment in my opinion but we haven't been consistent with it as such I don't feel strongly about it and we can leave the current name as it is for now.

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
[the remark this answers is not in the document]
DOBEN: I think it would be useful to see the output of the `concordium-client` at least once for the basic commands like `deploy/init/update` (these commands I expect not to change anytime soon). I leave it to you to decide if that adds value.

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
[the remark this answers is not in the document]
DOBEN: If we leave `ParseParam` then can we please update the comment here.

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
```diff
@@ -25,215 +25,211 @@ Once you have set up the tools needed you are ready to create your smart contrac
...
-    ) -> InitResult<State> {
-        // Your code
-
+    fn init(_ctx: &InitContext, _state_builder: &mut StateBuilder) -> InitResult<State> {
         Ok(State { counter: 0 })
     }
 
-Increment and decrement counter
+Update counter
 ===============================
```
DOBEN, line 59:
```suggestion
Update counter
============
```

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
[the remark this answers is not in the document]
DOBEN: Maybe:
```
Now we can invoke the update function with that input by using a contract update transaction. This will mutate the smart contract state and store the new value. If you have your own node, you can invoke the update function like so:
```

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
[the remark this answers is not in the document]
DOBEN: Yes, I don't think we need the ``--name`` option here (it can be removed since the name is not used anywhere in the tutorial).

If the ``--name MY_NAME`` option is set in the above command, you could use it:

```concordium-client contract init MY_NAME``` (no module reference needed)

DOBEN (approved):

source/mainnet/smart-contracts/tutorials/counter/counter-contract.rst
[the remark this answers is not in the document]
Victor-N-Suadicani: I can potentially add the output as text in the console code block, instead of as a screenshot - easier to update as well.

PR #1163, opened 2024-12-04
Fix redirect for buttons from the concordium website

dragosgd:
## Purpose

Fix redirect for links from buttons from the concordium website;
Also fixed 2 more links;

## Changes

problem was the following:

- when you use ./ is front, it's an absolute path, everything coming after https://developer.concordium.software/
- when not, it's relative, meaning everything after https://developer.concordium.software/en/mainnet/
- this goes both for the link you're redirecting from and the link you're redirecting to
- before, there was no need to differentiate as most paths were the similar, now, with the new structure, we need to see which need a relative path and which need an absolute path

## Checklist

      hard-to-understand areas.

## CLA acceptance

- [x] I accept the above linked CLA.


it09 (approved):
Thanks Dragos!

PR #1226, opened 2025-01-16
Added information about suspending a validator

TinaKT:
## Purpose

Add information about suspending validators relation to Protocol 8 changes.

## Changes

- Updated the page "Validation" with a section about suspension
- Added the term in the glossary

## Checklist

      hard-to-understand areas.

### DIFF
--- a/source/mainnet/docs/protocol/concepts-baker.rst
+++ b/source/mainnet/docs/protocol/concepts-baker.rst
@@ -20,6 +20,7 @@ A validator needs to :term:`stake<staked amount>` a part of its CCD balance on t
 
 To be chosen to produce a block, the validator must take part in a *lottery*. The greater the validator's stake, the greater the validator's chance of :term:`winning the lottery<lottery power>` and being selected to produce a block.
 
+
 Overview of the validation process
 ==================================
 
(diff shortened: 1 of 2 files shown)


source/mainnet/docs/protocol/concepts-baker.rst
```diff
@@ -104,3 +105,36 @@ A block is final when it cannot be rolled back any more, i.e., part of the autho
 
 A block is final at a minimum of two seconds after its creation. A new block has to be created descended from that block for the new block to be final.
 
+Validator Suspension
+====================
+
+A validator may be suspended if it remains inactive and fails to produce blocks when selected as a round leader. This mechanism, introduced in Protocol 8, helps maintain the blockchain's performance and reliability by excluding inactive validators.
+
+The :term:`suspension` process considers the validator's stake size:
```
td202, line 113: I think "considers" is not really accurate. Perhaps something like:

```suggestion
A validator faces :term:`suspension` if reaches a threshold of missed rounds. The size of a validator's stake affects how long the validator's node has to be inactive before reaching the threshold:
```

source/mainnet/docs/protocol/concepts-baker.rst
```diff
@@ -104,3 +105,36 @@ A block is final when it cannot be rolled back any more, i.e., part of the autho
...
+If the validator remains inactive through the snapshot epoch, the suspension takes effect at the following payday epoch.
+
+.. note::
+
+   A high-stake validator could face suspension after approximately 23 hours of inactivity, while low-stake validators typically require a longer period of inactivity before reaching the threshold.
+
+
+A suspended validator:
+
+* Must remain suspended for at least one epoch
```
td202, line 132: I think this line is probably more confusing than useful. It's not completely accurate - a validator can resume at any time after it is suspended. However, because the committee is updated at the snapshot epoch, when it is automatically suspended at a snapshot, it will be excluded from the committee for the next pay day. (If the validator is suspended manually, then it can be resumed even before the snapshot occurs.)

source/mainnet/docs/protocol/concepts-baker.rst
```diff
@@ -104,3 +105,36 @@ A block is final when it cannot be rolled back any more, i.e., part of the autho
...
+* Can resume validation through a manual transaction
+* Incurs no specific penalties beyond the suspension period
+* Maintains their delegators unless they choose to update their delegation
```
td202, line 135:
```suggestion
* Maintains their delegators unless the delegators choose to update their delegation
```

td202 (approved):

source/mainnet/docs/protocol/concepts-baker.rst
TinaKT: Deleted line

PR #1348, opened 2025-05-05
Update the documentation for the new version of the guardian app

soerenbf:
Adds documentation for the new feature revolving around (re)configuring the app to a new election target


chportma (approved):
It looks good. I have split the instructions about exporting the keys and submitting the first transaction into two different sections.

PR #1423, opened 2025-09-17
Fix cookie bug

dragosgd:
## Purpose

Fix cookie consent banner reappearing after users make their choice. The banner was showing up again on subsequent page visits even after users clicked "Allow" or "Decline", causing a poor user experience.

## Changes

Critical Bug Fixes:

- Fixed consentDenied() function which was incorrectly storing true instead of false when users declined cookies
- Replaced faulty consent checking logic that treated both granted and denied consent the same way

Improved Cookie Handling:

- Enhanced storeChoice() to set proper cookie attributes (1-year expiration, path scope, SameSite security)
- Updated loadChoice() to return null for no stored choice, distinguishing it from boolean consent values

Enhanced Logic Flow:

- Implemented three-way consent checking:

  - stored === true: Enable tracking and hide banner
  - stored === false: Hide banner without enabling tracking
  - stored === null: Show banner for first-time users

Safety Improvements:

- Added null checks before removing DOM elements to prevent JavaScript errors
- Only call Matomo's rememberCookieConsentGiven when consent was actually granted

## Checklist

      hard-to-understand areas.

## CLA acceptance

- [x] I accept the above linked CLA.


TinaKT (approved):
LGTM

